PR #25, opened 2023-05-29
[#24] Set up API routes 

### LINKED ISSUE #24: Set up API routes
## Why

Prepare the routes properly for future endpoints

## Acceptance Criteria

Make sure that all the endpoints start with `api/v1` (Update routes, controllers, ...)

## Design

N/A

## ResourcesN
N/A

Tuubz:
Close #24 

## What happened 👀

- Update `routes.rb` so all endpoints start with `api/v1`
- Add `health#show` as the root, to indicate the health status of the app
- Run test locally with `bundle exec rspec spec/requests/health_controller_spec.rb`
- Update `rubocop` to change the max amount of lines in a test to `6` 

## Insights ✍️ 

- Since `health_controller.rb` returns HTML, we must use `ActionController::Base`
- To fix the failing tests, update `codebase_spec.rb`
- [Your DB set-up must be set-up, before you can run a test](https://medium.com/geekculture/postgresql-rails-and-macos-16248ddcc8ba)
 
## Proof Of Work 📹

Screen Shot 2023-05-29 at 18 10 10

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class ApplicationController < ActionController::API
+class ApplicationController < ActionController::Base
   include Localization
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class ApplicationController < ActionController::Base
+class ApplicationController < ActionController::API
   include Localization
 end
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -11,20 +11,20 @@ class HealthController < ApplicationController
 
   def render_up
     respond_to do |format|
-      format.html { render html: html_status(color: 'green'), status: :ok }
+      format.html { render html: html_status('green'), status: :ok }
       format.json { render json: {}, status: :ok }
     end
   end
 
   def render_down(exception)
     respond_to do |format|
-      format.html { render html: html_status(color: 'red'), status: :internal_server_error }
+      format.html { render html: html_status('red'), status: :internal_server_error }
       format.json { render json: { error: exception.message }, status: :internal_server_error }
     end
   end
 
   # rubocop:disable Rails/OutputSafety
-  def html_status(color:)
+  def html_status(color)
     %(<html><body style="background-color: #{color}"></body></html>).html_safe
   end
   # rubocop:enable Rails/OutputSafety
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-class HealthController < ApplicationController
+# rubocop:disable Rails/ApplicationController
+class HealthController < ActionController::Base
   rescue_from Exception, with: :render_down
 
   def show
@@ -29,3 +30,4 @@ class HealthController < ApplicationController
   end
   # rubocop:enable Rails/OutputSafety
 end
+# rubocop:enable Rails/ApplicationController
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class HealthController < ApplicationController
+  rescue_from Exception, with: :render_down
+
+  def show
+    render_up
+  end
+
+  private
+
+  def render_up
+    respond_to do |format|
+      format.html { render html: html_status(color: 'green'), status: :ok }
+      format.json { render json: {}, status: :ok }
+    end
+  end
+
+  def render_down(exception)
+    respond_to do |format|
+      format.html { render html: html_status(color: 'red'), status: :internal_server_error }
+      format.json { render json: { error: exception.message }, status: :internal_server_error }
+    end
+  end
+
+  # rubocop:disable Rails/OutputSafety
+  def html_status(color:)
+    %(<html><body style="background-color: #{color}"></body></html>).html_safe
+  end
+  # rubocop:enable Rails/OutputSafety
+end
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -12,7 +12,7 @@ class HealthController < ApplicationController
   def render_up
     respond_to do |format|
       format.html { render html: html_status('green'), status: :ok }
-      format.json { render json: {}, status: :ok }
+      format.json { render json: { message: 'OK' }, status: :ok }
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,5 +4,6 @@ Rails.application.routes.draw do
       root "/health#show"
     end
   end
+  
   root "health#show"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
-  # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
-
-  # Defines the root path route ("/")
-  # root "articles#index"
+  namespace :api do
+    namespace :v1 do
+      # Defines the root path route ("/")
+      # root "articles#index"
+    end
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,7 @@
 Rails.application.routes.draw do
   namespace :api do
     namespace :v1 do
-      root "health#show"
+      root "/health#show"
     end
   end
   root "health#show"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,8 @@
 Rails.application.routes.draw do
   namespace :api do
     namespace :v1 do
-      # Defines the root path route ("/")
-      # root "articles#index"
+      root "health#show"
     end
   end
+  root "health#show"
 end
--- a/spec/codebase/codebase_spec.rb
+++ b/spec/codebase/codebase_spec.rb
@@ -23,14 +23,4 @@ describe 'Codebase', codebase: true do
 
     expect(invalid_engine_paths).to be_empty
   end
-
-  it 'does not contain respond_to blocks' do
-    find_results = `grep -r 'respond_to do' app/`
-    expect(find_results).to be_empty
-  end
-
-  it 'does not contain format blocks' do
-    find_results = `grep -r 'format.json' app/`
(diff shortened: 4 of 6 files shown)


config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
-  # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
-
-  # Defines the root path route ("/")
-  # root "articles#index"
+  namespace :api do
+    namespace :v1 do
+      # Defines the root path route ("/")
+      # root "articles#index"
+    end
```
olivierobert, line 6: A default route should be set up as a fallback (when no API routes is matched) and as a health endpoint (to check when the API is up or down):

```ruby
root "health#show"
```

Then:

```ruby
class HealthController < ApplicationController
  rescue_from(Exception) { render_down }

  def show
    render_up
  end

  private

  def render_up
    render html: html_status(color: "green")
  end

  def render_down
    render html: html_status(color: "red"), status: :internal_server_error
  end

  def html_status(color:)
    %(<html><body style="background-color: #{color}"></body></html>).html_safe
  end
end
```

Taken from https://github.com/rails/rails/pull/46936/files#diff-dfd63e49a9b67adb3b3ada9ac7e38ad7f29945b3f3db66eab4a663e98ac96ab1 that will come in Rails 7

Tuubz: @olivierobert Updated [here](https://github.com/nimblehq/ic-rails-huey-toby/pull/25/commits/24ae8461b7c04689dcf0c1aeeaab86c2b7654113)

> A default route should be set up as a fallback (when no API routes is matched) 

After the update, it will not fallback to root, from what I read the correct way is like this;

`match "*unmatched_route", to: "health#show" , via: :all`

Did I miss something? 🙏

olivierobert: You must be definitely doing something wrong as `root` mut work: https://guides.rubyonrails.org/routing.html#using-root

Maybe there is any issue with no routes being defined in the `api/v1` namespace. You might need to add a placeholder routes in there:

```ruby
namespace :api do
    namespace :v1 do
      	get '/', to: 'app#index'
    end
end
```

Tuubz: @olivierobert 

`root` is showing correctly, once we navigate to `http://0.0.0.0:3000/`

Do I understand correctly that  `http://0.0.0.0:3000/invalid` should navigate to `root` too? 💭

app/controllers/health_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def render_down
+    render html: html_status(color: 'red'), status: :internal_server_error
+  end
+
+  # rubocop:disable Rails/OutputSafety
+  def html_status(color:)
+    %(<html><body style="background-color: #{color}"></body></html>).html_safe
+  end
```
hoangnguyen92dn, line 23: @Tuubz Since we are developing API application so do you think we should return `json` as default instead? 😄 

```
    def render_up
        render json: { status: :ok }
    end

    def render_down
      render json: { status: :internal_server_error }
    end
```

olivierobert: That's not for API consumers but you can make use of `respond_to` with different formats (see example: https://guides.rubyonrails.org/action_controller_overview.html#rendering-xml-and-json-data):

```
def show
  respond_to do |format|
      format.html  { render_up }
      format.json { render json: {}, status: :ok}
    end
end
```

hoangnguyen92dn: @olivierobert I tried with `respond_to` to support `html` and `json`, it works on the browser when I visit `localhost:3000`, when trying on Postman it always returns the response as `html` format. I found a suggestion about adding `.json` suffix to the endpoint but not sure how it work with the root. Am I missing something? 🤔 

image

image

olivierobert: I believe you are missing a second header `Accept: application/json` in addition to `Content-Type`

See https://webmasters.stackexchange.com/questions/31212/difference-between-the-accept-and-content-type-http-headers

Welcome to the Web and its intricacies 🔬

hoangnguyen92dn: @olivierobert Niceeee! 🎉 Finally it works 🚀 cc @Tuubz

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
-  # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
-
-  # Defines the root path route ("/")
-  # root "articles#index"
+  namespace :api do
+    namespace :v1 do
+      root "/health#show"
```
olivierobert, line 4: It is kind of repetitive though 🤔 And there should be only one `root`, no? Will this be removed?

config/routes.rb
```diff
@@ -1,6 +1,8 @@
...
+    end
+  end
```
olivierobert, line 6: A line break between each route block would be better for readability

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class ApplicationController < ActionController::API
+class ApplicationController < ActionController::Base
```
olivierobert, line 3: This change is not really needed IMO. It was fine as it was.

olivierobert (commented):
It would be best if you also implemented `request` tests to verify the implementation. It looks like the following:

image

app/controllers/application_controller.rb
hoangmirs: As I remember, the `ActionController::API` doesn't have `respond_to` method. But I agree that we should not change this as this is an API application.
Should we just let the `HealthController` inherits `ActionController::Base` only?

config/routes.rb
Tuubz: Since we don't want `baseUrl/api/v1` to break, I've added a root here 🙏

spec/requests/health_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
+RSpec.describe 'Health Controller', type: :request do
```
hoangnguyen92dn, line 3: I am not sure about the test class name and also the controller test, following [Compass](https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#convention-13) we don't write controller test at all 🤔

spec/codebase/codebase_spec.rb
```diff
@@ -23,14 +23,4 @@
 
     expect(invalid_engine_paths).to be_empty
   end
-
-  it 'does not contain respond_to blocks' do
```
hoangnguyen92dn, line 27: Just curious...For the web application, we should avoid using `respond_to` and `format.json`? 🤔

olivierobert: No, it is fine to use :-) It is just that normally we recommend rendering a view not just some HTML string directly.

spec/requests/health_controller_spec.rb
olivierobert: Normally, we do not write many controller tests. We prefer writing request tests or UI tests.

So since it is an API, we can name it like the following:

```suggestion
RSpec.describe 'Health Endpoint', type: :request do
```

spec/requests/health_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  describe 'GET #show' do
+    context 'when the application is up' do
+      it 'returns a successful response with HTML' do
+        get '/'
```
olivierobert, line 7: Always add aline break between each block in the Arrange, Act and Assert (AAA) Pattern:

```suggestion
        get '/'
        
```

spec/requests/health_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      it 'returns a successful response with JSON' do
+        get '/', headers: { 'ACCEPT' => 'application/json' }
+        expect(response).to have_http_status(:ok)
+        expect(response.content_type).to eq('application/json; charset=utf-8')
+        expect(response.body).to eq({ message: 'OK' }.to_json)
+      end
+    end
+
+    context 'when the application is down' do
+      before(:each) do
```
olivierobert, line 22: Usage of `before` blocks in Rspec should be avoided: https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#convention-15

There is a lot of complexity under the hood

spec/requests/health_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        error = StandardError.new('Something went wrong')
+        allow_any_instance_of(HealthController).to receive(:show).and_raise(error)
```
olivierobert, line 24: `allow_any_instance_of` is a dirty way to achieve this 🙈 See a better mocking approach: https://stackoverflow.com/a/27144891

olivierobert (changes requested):

spec/codebase/codebase_spec.rb
Tuubz: @hoangnguyen92dn For API application, there should not be any `respond_to`

But since we added the health page, I had to remove it 🙏

spec/requests/health_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        get '/', headers: { 'ACCEPT' => 'application/json' }
+
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to eq({ message: 'OK' }.to_json)
+      end
+    end
+
+    context 'when the application is down' do
+      it 'returns an internal server error response with HTML' do
+        controller_instance = HealthController.new
```
Tuubz, line 25: Initially, I tried with `controller_instance = instance_double(HealthController)`

But this would result to, not sure if if I'm doing this correctly 💭 

```
Failure/Error: get '/'
#<InstanceDouble(HealthController) (anonymous)> received unexpected message :dispatch with ("show", #<ActionDispatch::Request GET "http://www.example.com/" for 127.0.0.1>, #<ActionDispatch::Response:0x00007f8d072ad5a0 @mon_data=#<Monitor:0x00007f8d072ad2f8>, @mon_data_owne... @cache_control={}, @request=#<ActionDispatch::Request GET "http://www.example.com/" for 127.0.0.1>>)
```

olivierobert: Controllers are a bit of a pain in Rails so let's not worry too much about it in this case 😌

olivierobert (approved):

hoangnguyen92dn (approved):

Issue #19, opened 2023-05-26
[Backend] As a user, I can run a background job for processing CSV file with keywords


Tuubz:
## Why

A user can run background job to process the uploaded CSV file with keywords

## Acceptance Criteria

- Add a new field to the `search_results` table, with the field `status` (in progress, completed or failed)
- Start a background job and update the status accordingly (don't start scraping yet)
- Update the status accordingly 
- Write unit and integration tests

**Note**: Make use of `sidekiq` to handle the background job

## Design
N/A

## Resources
https://github.com/sidekiq/sidekiq

![Image](https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/8c4713db-262c-46be-8b35-16d61dc79a56)

Issue #15, opened 2023-05-25
[Backend] As a user, I can get scraped information from a Google search page


Tuubz:
## Why

Scrape the details from the search result for every keyword in the CSV file (from Google)

## Acceptance Criteria

- Scrape and store the following details as a result, from the raw HTML for every keyword:
    - Number of AdWords advertisers in the top position.
    - Total number of AdWords advertisers on the page.
    - URLs of the AdWords advertisers in the top position. (array to string)
    - Number of the non-AdWords results on the page.
    - URLs of the non-AdWords results on the page. (array to string)
    - Total number of links (all of them) on the page 
    - HTML code of the page/cache of the page. (#14)
    - Search engine used (Google or Bing)

- Write unit and integration tests

**Note**:  Use `nokogiri` to scrape the raw HTML

## Design
N/A

## Resources
https://github.com/sparklemotion/nokogiri

![Image](https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/8c4713db-262c-46be-8b35-16d61dc79a56)

PR #30, opened 2023-05-31
[#14] [Backend] As a user, I can get the raw HTML of one Google search page

### LINKED ISSUE #14: [Backend] As a user, I can get the raw HTML of one Google search page
## Why

A search result must be made for one keyword in the CSV file, once the user uploads the file (Google)

## Acceptance Criteria

- Make sure we're able to scrape the raw HTML for _one_ keyword (and store it in the `html_code` field):

## Design
N/A

## Resources

hoangnguyen92dn:
- Close #14 

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


app/services/search/google/search_service.rb
```diff
@@ -0,0 +1,21 @@
...
+      GOOGLE_SEARCH_BASE_URL = 'https://www.google.com/search'
+
+      def initialize(keyword, language = 'en')
+        super(keyword, language)
+      end
+
+      def search
+        uri = URI.parse("#{GOOGLE_SEARCH_BASE_URL}?q=#{keyword}&hl=#{language}&lr=#{language}")
+
+        response = Net::HTTP.get_response(uri)
```
Tuubz, line 15: Just curious, any reason we didn't use Nokogiri here? 🙏 

It will be beneficial once we start scraping the actual HTML

db/migrate/20230601025347_create_search_results.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :keyword
+      t.integer :status
+      t.integer :adwords_top_count
+      t.integer :adwords_total_count
+      t.text :adwords_top_urls
+      t.integer :non_adwords_count
+      t.text :non_adwords_urls
+      t.integer :total_links_count
+      t.text :html_code
+      t.string :search_engine
```
Tuubz, line 13: Should we only add the fields that we're using in this PR?

We can always update the table later on  🚀

app/models/search_result.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class SearchResult < ApplicationRecord
+  enum status: { processing: 0, completed: 1 }
+end
```
Tuubz, line 5: Should this be implemented in #19 instead? 🙏

app/services/search/google/search_service.rb
hoangnguyen92dn: @Tuubz `Nokogiri` is supposed for parsing HTML content. With the feature scope, we won't parse any the HTML to get the values so we don't need to use it.

db/migrate/20230601025347_create_search_results.rb
hoangnguyen92dn: @Tuubz 👌 Then we will need to clean the database to avoid invalid old records in the database.

app/services/search/google/search_service.rb
Tuubz: You are correct that we won't parse the HTML content yet.

But if I'm not mistaken we can make get the response with Nokigiri, to then later parse the HTML later on

```
require 'nokogiri'
require 'open-uri'

# Fetch and parse HTML document
doc = Nokogiri::HTML(URI.open('https://nokogiri.org/tutorials/installing_nokogiri.html'))

# Search for nodes by css
doc.css('nav ul.menu li a', 'article h2').each do |link|
  puts link.content
end
```

Reference: https://github.com/sparklemotion/nokogiri#parsing-and-querying

Tuubz: Oh, I see my mistake now.

So in the next issue we can call `Nokogiri::HTML(response)`, right? 🙏

hoangnguyen92dn: > So in the https://github.com/nimblehq/ic-rails-huey-toby/issues/15 we can call Nokogiri::HTML(response.body), right? 🙏

Yup 👍

app/models/search_result.rb
hoangnguyen92dn: @Tuubz 🙆‍♂️

db/migrate/20230601025347_create_search_results.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSearchResults < ActiveRecord::Migration[7.0]
+  def change
+    create_table :search_results do |t|
+      t.string :keyword
+      t.text :html_code
```
hoangmirs, line 5: In the AC, I saw an `Update the status accordingly` requirement, so should we add it to this PR as well? 🤔

spec/services/search/google/search_service_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Search::Google::SearchService, type: :service do
+  describe '#call' do
+    context 'given the search result is valid' do
+      it 'returns search page body', vcr: 'search_service/keyword_ruby' do
```
hoangmirs, line 8: You can name the VCR as `services/search/google/valid` to know which current file is used for which test

spec/services/search/google/search_service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+        search_result = described_class.new('ruby').call
+
+        expect(search_result).to include('<title>ruby - Google Search</title>')
+      end
+    end
+
+    context 'given the search result is invalid' do
```
hoangmirs, line 15: Sorry, this might not be on the Compass, but for invalid cases, we often uppercase the `INVALID` text:
```suggestion
    context 'given an INVALID search result' do
```

app/services/search/google/search_service.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Google
+    class SearchService < Search::BaseService
+      GOOGLE_SEARCH_BASE_URL = 'https://www.google.com/search'
+
+      def initialize(keyword, language = 'en')
+        super(keyword, language)
+      end
+
+      def call
+        uri = URI.parse("#{GOOGLE_SEARCH_BASE_URL}?q=#{keyword}&hl=#{language}&lr=#{language}")
```
hoangmirs, line 13: As you're using a base service, should we default a method `search_url` that uses `BASE_URL` constant and gets the corresponding value?

db/migrate/20230601025347_create_search_results.rb
Tuubz: My bad, I think it would be better to keep it as part of this issue: #19 🙏

hoangnguyen92dn: @hoangmirs As discussed with Toby, the status should be updated accordingly when the search job starts and the scraping data is completed 🙏

spec/services/search/google/search_service_spec.rb
hoangmirs: No, we usually apply it for words like `invalid`/`no` to emphasize that these are unusual cases 🙌

hoangnguyen92dn: @hoangmirs 🙆‍♂️  Thanks for clarifying 🚀

app/services/search/google/search_service.rb
hoangnguyen92dn: @hoangmirs The base search service has been refactored 🚀

spec/support/vcr.rb
```diff
@@ -10,7 +10,7 @@
   c.ignore_request do |request|
     URI(request.uri).port == 9200
   end
-  c.default_cassette_options = { record: :none, match_requests_on: [:path] }
+  c.default_cassette_options = { record: :once, match_requests_on: [:path] }
```
Tuubz, line 13: Just curious, why was this change necessary? 🙏

spec/services/search/google/search_service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      it 'returns search page body', vcr: 'services/search/google/valid' do
+        search_result = described_class.new('ruby').call
+
+        expect(search_result).to include('<title>ruby - Google Search</title>')
+      end
+    end
+
+    context 'given an INVALID search result' do
+      it 'returns nil' do
+        stub_request(:any, /www\.google\.com/).to_return(status: 404, body: '', headers: {})
```
Tuubz, line 17: Optional:
```suggestion
        stub_request(:any, /www\.google\.com/).to_return(status: 404)
```

spec/fixtures/vcr/services/search/google/valid.yml
Tuubz, line 1: Curious again, lol

How did you get this? 🙏

spec/support/vcr.rb
hoangnguyen92dn: @Tuubz It means that if the vcr file is not available, we will need to record (cache) the request and response at least 1 time for the next execution. (Ref: https://github.com/vcr/vcr#usage)

spec/fixtures/vcr/services/search/google/valid.yml
hoangnguyen92dn: @Tuubz This file is recorded by [vcr](https://github.com/vcr/vcr#vcr) 😄 In the test case we define like this, unlike the integration test in the mobile app, in rails application it will send the real request and we use vcr to cache the response for the next usage 😄 

```
it 'returns search page body', vcr: 'services/search/google/valid' do
```

spec/support/vcr.rb
Tuubz: > It means that if the vcr file is not available

You mean like removing `vcr: 'services/search/google/valid'`, right?

I was expecting it to record the request after removing it, but it doesn't 

Please let me know if I understood incorrectly 🙇

spec/fixtures/vcr/services/search/google/valid.yml
Tuubz: @hoangnguyen92dn I see, but what do I need to do to generate `valid.yml`? 🙇

hoangnguyen92dn: @Tuubz When you execute the test, it will record and generate the file automatically.

Tuubz (approved):

spec/support/vcr.rb
hoangnguyen92dn: > I was expecting it to record the request after removing this from the test, but it doesn't

It will generate the file, that's why we have to change to `:once` 😄

app/services/search/base_service.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Search
+  class BaseService
```
olivierobert, line 4: This implantation does not follow idiomatic Ruby 😅 This more like Java or Kotlin OOP style.

A more Ruby-ish way is to use the factory pattern:

```suggestion
  class SearchService
    def initialize(search_engine:)
      case 'google'
        Search::Google.new(keyword, language
      case 'bing'
        Search::Bing.new(keyword, language
      else
        raise NotImplementedError
      end
    end
  end
```

app/services/search/base_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    def initialize(search_url)
+      @search_url = search_url
+    end
+
+    def call
+      uri = URI.parse(search_url)
+
+      response = Net::HTTP.get_response(uri)
+
+      response.body if response.is_a?(Net::HTTPSuccess)
```
olivierobert, line 14: There is no error handling in case of error 😬

app/services/search/base_service.rb
```diff
@@ -0,0 +1,21 @@
...
+module Search
+  class BaseService
+    def initialize(search_url)
+      @search_url = search_url
+    end
+
+    def call
+      uri = URI.parse(search_url)
+
+      response = Net::HTTP.get_response(uri)
```
olivierobert, line 12: It might be better to use Faraday or Httparty as this standard library is quite poor (especially to handle errors and using different user agents, the latter will be needed)

app/services/search/google/search_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Search
+  module Google
+    class SearchService < Search::BaseService
+      def initialize(keyword, language = 'en')
+        search_url = "https://www.google.com/search?q=#{keyword}&hl=#{language}&lr=#{language}"
```
olivierobert, line 7: Never hard code such values. The static parts of the URL should be defined as constants.

olivierobert (changes requested):

app/services/search/base_service.rb
Tuubz: I'll take care of this in #19, as it's required to set the status to `:failed` when there's an error 🙏

Tuubz: Let's leave a TODO message, just to be sure I don't forget 🙈

app/services/search/base_service.rb
hoangnguyen92dn: By refactoring to apply the factory pattern, I moved the search function to `GoogleSearchService` class, the same will be added for BingSearchService.

app/services/search/google_search_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def search
+      uri = URI.parse(@search_url)
+
+      response = Faraday.get(uri)
+
+      # TODO: Handle error in background job
+      # https://github.com/nimblehq/ic-rails-huey-toby/issues/19
+      response.body if response.status == 200
```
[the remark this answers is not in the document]
Tuubz: @olivierobert @hoangmirs 

FYI: I'll store the `response_body` (html_code) when working on #19, as it should be updated in the job 🙏

db/migrate/20230530101854_create_search_results.rb
```diff
@@ -3,6 +3,7 @@ def change
     create_table :search_results do |t|
       t.string :keyword
       t.string :search_engine
+      t.text :html_code
```
Tuubz, line 6: Normally we should never update already existing migrations, it's a better practice to create a new migration.

Imagine we have a lot of data in production, then we would need to drop first in order to migrate successfully.

But for this case, it's OK, just something to keep in mind 🙏 

@olivierobert @hoangmirs Please let me know if I'm wrong 🙇

Tuubz (commented):
LGTM 👍

Tuubz (commented):
LGTM 👍

app/services/search/google_search_service.rb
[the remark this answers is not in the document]
Tuubz: @hoangmirs @olivierobert 

I'll store `response.body` (html_code`) in #19, as it should be updated in the job 🙏

db/migrate/20230530101854_create_search_results.rb
hoangnguyen92dn: @Tuubz Good catch 👍 New migration file pushed 🚀 [f71716e](https://github.com/nimblehq/ic-rails-huey-toby/pull/30/commits/f71716ec795176418fe7e320578829d65d21e4c6)

spec/services/search/search_service_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Search::SearchService, type: :service do
+  describe '#initialize' do
```
olivierobert, line 6: For a class methods, it should be:

```suggestion
  describe '.new' do
```

See https://www.betterspecs.org/#describe

spec/services/search/search_service_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    context 'given the search engine is INVALID' do
+      it 'raises an error' do
+        expect { described_class.new(search_engine: 'yahoo', keyword: 'ruby') }.to raise_error(NotImplementedError)
+      end
+    end
+  end
+
+  describe '#search' do
```
olivierobert, line 21: Shouldn't this test be defined in another test file for `GoogleSearchService`? That instance method does not exist on `GoogleSearchService`.

app/services/search/search_service.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Search
+  class SearchService
+    def self.new(search_engine:, keyword:, language: 'en')
+      case search_engine
+      when 'google'
```
hoangmirs, line 7: We use `symbol` over `string` for these cases
```suggestion
      when :google
```

hoangmirs (approved):

app/services/search/search_service.rb
hoangnguyen92dn: @hoangmirs 👍 Thank you. Good to know.

Tuubz:
@olivierobert One more approval/review is needed here 🚀

olivierobert (approved):

PR #31, opened 2023-06-01
[#12] [Backend] As a user, I can upload a CSV file with keywords #12

### LINKED ISSUE #12: [Backend] As a user, I can upload a CSV file with keywords
## Why

A table row must be created for every keyword in the CSV file.

## Acceptance Criteria

- Create a new table for keywords, besides the default fields, add a name field.
- Add references with the users table (one user has many keywords & one keyword has one user)
- Create a new endpoint: `POST ap/v1/upload?engine={google/bing}`
    - **Request**: CSV file
    - **Response**: Success message
- Start one background job that processes the keywords into the table.

**Note**: Make use of `sidekiq` to handle the background job

## Design
N/A

## Resources
- https://www.postman.com/postman/workspace/postman-answers/documentation/13455110-00378d5c-5b08-4813-98da-bc47a2e6021d
- https://github.com/sidekiq/sidekiq

Tuubz:
Close #12

## What happened 👀

- Created search results table, which has a keyword and a search engine (Google or Bing)
- Add route, controller, validation, localizations, and tests

## Insight 📝

Instead of creating a new row one by one, it's more performant to add them all at once.

We do this by using [activerecord-import](https://github.com/zdennis/activerecord-import), because rubucop doesn't like `insert_all` 🚀 

---

Add `active_record_enum.rb` to solve this issue: https://nimble-co.slack.com/archives/C058GEPCT7F/p1686033730579809

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/113abd80-83b1-4345-a993-4e71e0808fa9


app/models/search_result.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SearchResult < ApplicationRecord
+  SEARCH_ENGINES = %w[google bing].freeze
```
Tuubz, line 4: Trying to use enum here instead 💭

Gemfile
```diff
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 ruby '3.0.1'
 
 # Backend
+gem 'activerecord-import'
```
olivierobert, line 5: Always add a comment to each gem to provide information on why the gem was added (see the either gems for example)

app/controllers/api/v1/upload_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+require 'csv'
+
+module Api
+  module V1
+    class UploadController < ApplicationController
+      def create
+        upload = Upload.new(
+          search_engine: params[:search_engine],
+          csv_file: params[:csv_file]
```
olivierobert, line 11: Prefer to use strong params: https://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html

So there should be a private method to permit specific params

app/controllers/api/v1/upload_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+        )
+        if upload.save
+          render json: { message: I18n.t('upload.upload_success'), status: :ok }
```
olivierobert, line 14: 1. It would be better to follow [JSON API spec](https://jsonapi.org/) so with `data` attribute. 
2. We generally use serializer to render API responses (`app/serializers`). Past project can provide an example.
3. When creating data, the status code should be `201`: http://www.railsstatuscodes.com/created.html#:~:text=201%20%2D%20Created&text=The%20request%20has%20been%20fulfilled,a%20new%20resource%20being%20created.

app/models/upload.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def save
+    return false unless valid?
+
+    begin
+      content = csv_file.read
+      keywords = CSV.parse(content).flatten.map(&:strip)
+      search_results = keywords.map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
+      SearchResult.import search_results, validate: true, raise_error: true
```
olivierobert, line 24: Let's made this code breadth a bit for redability:

```suggestion
      content = csv_file.read
      keywords = CSV.parse(content).flatten.map(&:strip)
    
      search_results = keywords.map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
      
      SearchResult.import search_results, validate: true, raise_error: true
```

app/models/upload.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+gem 'activerecord-import'
+
+class Upload
+  include ActiveModel::Validations
```
olivierobert, line 6: Good idea to use `ActiveModel` 👍

app/models/search_result.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SearchResult < ApplicationRecord
+  enum search_engines: { google: 0, bing: 1 }
```
olivierobert, line 4: Also prefer to keep naming to singular to avoid typos and guesses:

```suggestion
  enum search_engine: { google: 0, bing: 1 }
```

FYI there are two syntax you pick from: https://blog.saeloun.com/2021/02/26/rails-introduces-new-syntax-for-enum/#after

app/models/search_result.rb
```diff
@@ -0,0 +1,8 @@
...
+
+  validates :keyword, presence: true
+  validates :search_engine, presence: true
```
olivierobert, line 7: We might want to validate that `search_engine` is one of the valid one. Even though `UploadValidator` has some validation for it, for data integrity, the model must have such validation, i.e., a developer could insert new data in that table without passing through the `Upload` mode.

app/models/upload.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+gem 'activerecord-import'
+
+class Upload
```
olivierobert, line 5: This could be a form object (see https://nimblehq.co/blog/lets-play-design-patterns-form-objects), since it is called in the controller https://github.com/nimblehq/ic-rails-huey-toby/pull/31/files#diff-1faaea254bbc4f61853695370bce10e777fc843b909b579a983b8d417072bb6aR9

app/validators/upload_validator.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def search_engine
+    upload.search_engine
+  end
+
+  def csv_file
+    upload.csv_file
+  end
+
+  def validate_search_engine
```
olivierobert, line 25: All these validations could be defined in https://github.com/nimblehq/ic-rails-huey-toby/pull/31/files#diff-becd1596dd6a914113c0a2b7b5cde93d201b7bfeb4c5a9ca27b96950ce329339R5 using `ActiveModel::Validations` DSL. Usually, we write a custom, validator for 1) a single validation (not multiple) and 2) for validations that we cannot do with the built-in DSL.

config/locales/en.yml
```diff
@@ -31,3 +31,12 @@
...
   hello: "Hello world"
+  upload:
+    upload_success: "Upload success."
+    validation:
+      save_failed: "Failed to save search results."
+      search_engine_empty: "Search engine can't be empty: search_engine."
+      search_engine_invalid: "Search engine is invalid, must be 'google' or 'bing'."
+      csv_file_empty: "CSV file can't be empty: csv_file."
+      csv_file_invalid_extension: "CSV file is invalid, must have .csv extension."
+      csv_file_invalid_count: "CSV file is invalid, must have 1-1000 keywords."
```
olivierobert, line 42: Rails has a built-in way to handle model validation errors(see: https://stackoverflow.com/posts/40604185/revisions)

spec/fixtures/files/upload_valid.csv
```diff
@@ -0,0 +1,6 @@
+vpn
+smartphone
+internet
+marketing
+iphone
+samsung
```
olivierobert, line 6: You might want to try that it works with keywords compose of multiple words:

```suggestion
best samsung phones 2023
```

spec/models/search_result_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require 'rails_helper'
+
+RSpec.describe SearchResult, type: :model do
+  describe 'validations' do
+    it { is_expected.to validate_presence_of(:keyword) }
+    it { is_expected.to validate_presence_of(:search_engine) }
+  end
+
+  describe 'error message' do
```
olivierobert, line 11: This is still part of validations, no?

spec/models/upload_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          csv_file: fixture_file_upload('upload_invalid_count.csv', 'text/csv')
+        )
+
+        expect(upload.save).to be(false)
+        expect(upload.errors.full_messages).to eq([I18n.t('upload.validation.csv_file_invalid_count')])
+        expect(SearchResult.count).to eq(0)
+      end
+    end
+
+    context 'given csv file can not be read' do
```
olivierobert, line 72: Minor english issue:

```suggestion
    context 'given a CSV file cannot be read' do
```

Usually for sad path we prefer to highlight `NOT`:

```suggestion
    context 'given the CSV file is NOT readable' do
```

app/models/upload.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def save
+    return false unless valid?
+
+    begin
+      content = csv_file.read
+      keywords = CSV.parse(content).flatten.map(&:strip)
+      search_results = keywords.map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
+      SearchResult.import search_results, validate: true, raise_error: true
+    rescue StandardError
```
olivierobert, line 25: Is there not a specific custom error thrown instead of catching with `StandardError`? The more specific, the better,

olivierobert (changes requested):

app/models/upload.rb
Tuubz: @olivierobert Is it possible that the suggested change is incorrect? 🙏

app/controllers/api/v1/upload_controller.rb
Tuubz: Since we're returning a simple success message, is it necessary to follow step 1 & 2? 💭

app/models/upload.rb
olivierobert: Yes, the suggestion deals with all the lines of code above. It is meant more as an example of how the whole code should be structured.

app/controllers/api/v1/upload_controller.rb
olivierobert: On creation, the API should return a success message and the created resources. It is a best practice. So yes following the steps 1 and 2 are needed.

app/controllers/api/v1/upload_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Api
+  module V1
+    class UploadController < ApplicationController
+      def create
+        upload = Upload.new(
```
hoangmirs, line 9: You can change this object into a [Form object](https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#forms)

app/controllers/api/v1/upload_controller.rb
hoangmirs: Since you're processing bulk import, you can return the total keywords in the response 💭

app/controllers/api/v1/upload_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'csv'
```
hoangmirs, line 3: I think we don't need this as Rails autoloaded it already

app/controllers/api/v1/upload_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Api
+  module V1
+    class UploadController < ApplicationController
```
hoangmirs, line 7: In Rails, controller names should be in plural form. This convention is followed to maintain consistency and to make it easier for developers to understand the code of a Rails application. In this case, it is `UploadsController` to create for `Upload` resource. But as this controller is for creating `SearchResults`, I think it should be `SearchResultsController`

db/migrate/20230530101854_create_search_results.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSearchResults < ActiveRecord::Migration[7.0]
+  def change
+    create_table :search_results do |t|
+      t.string :keyword
+      t.integer :search_engine
```
hoangnguyen92dn, line 5: Please update the data type if we consider to change the value of enum for `search_engine`

app/controllers/api/v1/upload_controller.rb
Tuubz: Sadly, it doesn't work if I remove it: `uninitialized constant UploadForm::CSV`

Tuubz:
Updated @olivierobert @hoangmirs @hoangnguyen92dn 🥴

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SearchResultsController < ApplicationController
```
hoangnguyen92dn, line 5: @Tuubz By renaming the controller, the upload will belong to `search_results` endpoint?  `api/v1/search_results/upload` 🤔

Tuubz: No, only `http://0.0.0.0:3000/api/v1/upload` will be supported 🚀

hoangnguyen92dn: Ah ha...There is `resources :upload` 🆗

hoangnguyen92dn (approved):

app/forms/upload_form.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def valid_content_type
+    return if csv_file.content_type == 'text/csv'
+
+    add_error :csv_file_invalid_extension
+  end
+
+  def valid_count
+    return if CSV.read(csv_file.tempfile).count.between?(1, 1000)
```
olivierobert, line 29: The Mac count should be 100 and the value should be defined as a constant

config/locales/en.yml
olivierobert, line 1: These localizations do not follow the conventions I recommended before. It needs to be in a file named activemodel.en.yml. It works the same as https://github.com/nimblehq/mia-web/blob/development/config/locales/activerecord.en.yml

spec/forms/upload_form_spec.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe UploadForm, type: :model do
+  describe '#valid?' do
+    context 'when all attributes are valid' do
```
olivierobert, line 7: Prefer the formulation as the Web chapter uses it more:

```suggestion
    context 'given valid all attributes' do
```

spec/requests/search_results_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Search Results Endpoint', type: :request do
```
olivierobert, line 5: It can be simplified to:

```suggestion
RSpec.describe 'Search Results', type: :request do
```

spec/fixtures/files/upload_invalid_extension.txt
```diff
@@ -0,0 +1 @@
+keyword
```
olivierobert, line 1: Missing end of file line break

app/forms/upload_form.rb
```diff
@@ -0,0 +1,47 @@
...
+  def add_error(type)
+    errors.add(:base, I18n.t("upload_form.validation.#{type}"))
+  end
+
+  def save
+    content = csv_file.read
+    keywords = CSV.parse(content).flatten.map(&:strip)
+
+    search_results = keywords.map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
+    SearchResult.import search_results
```
olivierobert, line 43: Why using import over just the save method each keyword result?

olivierobert (changes requested):

app/forms/upload_form.rb
Tuubz: I'll move it as a constant, however the requirements specify 1000 (not 100)

>  The uploaded file can be of any size from 1 to 1,000 keywords.

app/forms/upload_form.rb
Tuubz: From my understanding, this approach is more performant

Ex: 100 keywords, instead of doing 100 database transactions, we only do one transaction 🚀

olivierobert: You are right but prefer to use the built-in [insert_all](https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/insert_all) method provided by ActiveRecord

olivierobert (changes requested):
Just one minor suggestion for improvements and then it would be good to go

app/forms/upload_form.rb
Tuubz: Oh, I initially did, but rubocop gave a warning about it, should I still proceed? 💭

Tuubz: Oh, I initially did, but rubocop gave a warning about it, should I still proceed? 💭

Tuubz: I'll give it another try

olivierobert: Oh I see then that should be okay to move forward with the current implementation

olivierobert (approved):

app/forms/upload_form.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def valid_count
+    return if CSV.read(csv_file.tempfile).count.between?(CSV_COUNT_MIN, CSV_COUNT_MAX)
+
+    add_error :invalid_count
+  end
+
+  def add_error(type)
+    errors.add(:base, I18n.t("activemodel.errors.models.search_result.attributes.csv_file.#{type}"))
+  end
```
hoangmirs, line 40: You don't need to get the error message with this long line. To add a custom error to an attribute, you just need:
```suggestion
  def add_error(type)
    errors.add(:csv_file, type)
  end
```
And in the `activemodel.yml`:
```yml
en:
  activemodel:
    errors:
      models:
        upload_form:
          attributes:
            csv_file:
              blank: "CSV file is required."
              invalid_extension: "CSV file is invalid, must have .csv extension."
              invalid_count: "CSV file is invalid, must have 1-1000 keywords."
```

app/forms/upload_form.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def valid_content_type
+    return if csv_file.content_type == CSV_CONTENT_TYPE
+
+    add_error :invalid_extension
+  end
+
+  def valid_count
+    return if CSV.read(csv_file.tempfile).count.between?(CSV_COUNT_MIN, CSV_COUNT_MAX)
```
hoangmirs, line 33: This might raise an error when you pass a non-CSV file. We should validate this when the `csv_file` is a CSV already

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      def create
+        upload_form = UploadForm.new(
+          search_engine: upload_form_params[:search_engine],
+          csv_file: upload_form_params[:csv_file]
+        )
+        if upload_form.valid?
```
hoangmirs, line 11: Please add a break line between blocks 🙏

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      private
+
+      def upload_form_params
+        params.permit(:search_engine, :csv_file)
+      end
+
+      def render_success(search_results)
+        data = SearchResultsSerializer.new(search_results).serializable_hash[:data]
+        message = I18n.t('activemodel.notices.models.search_result.create')
+        render json: { data: data, meta: { message: message } }, status: :created
```
hoangmirs, line 28: We should add a new line before the render method 🙏

spec/models/search_result_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe SearchResult, type: :model do
+  describe 'validations' do
+    it { is_expected.to validate_presence_of(:keyword) }
+
+    it { is_expected.to validate_presence_of(:search_engine) }
+
+    it 'is invalid without a keyword' do
+      search_result = described_class.new(keyword: nil)
+      search_result.valid?
+      expect(search_result.errors[:keyword]).to include("can't be blank")
```
hoangmirs, line 14: Please add space to separate blocks:
```suggestion
      search_result = described_class.new(keyword: nil)

      search_result.valid?

      expect(search_result.errors[:keyword]).to include("can't be blank")
```

Tuubz:
Thank you @hoangmirs , I'll add your changes in my next PR 🙏

PR #33, opened 2023-06-07
[#19] [Backend] As a user, I can run a background job for processing CSV file with keywords

### LINKED ISSUE #19: [Backend] As a user, I can run a background job for processing CSV file with keywords

Tuubz:
Close #19

## What happened 👀

- Add [sidekiq](https://github.com/sidekiq/sidekiq) (background job handler), with a route to access the [dashboard](http://localhost:3000/sidekiq)
- Add status (`in_progress`, `completed` or `failed`) for SearchResult (enum)
- Run background job and update `html_code` & `status` for the search result accordingly
- Add tests

## Insight 📝

To run sidekiq locally, you must execute: `bundle exec sidekiq `

This will not only execute queued jobs, but it will also show the logs. Optionally, we can observe the[ dashboard](http://localhost:3000/sidekiq)

Since our app is an API app (`config.api_only = true`), we must add the following to support the dashboard:
```
# This is required for an API app to enable the Sidekiq dashboard
config.session_store :cookie_store, key: '_interslice_session'
config.middleware.use ActionDispatch::Cookies
config.middleware.use config.session_store, config.session_options
```

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/502e431b-b7b8-49f0-a062-1fd94a908115


app/controllers/api/v1/search_results_controller.rb
```diff
@@ -21,6 +19,23 @@ def create
       def upload_form_params
         params.permit(:search_engine, :csv_file)
       end
+
+      def process_results(search_results)
+        search_results.each do |search_result|
+          SearchKeywordJob.perform_later(search_result)
```
olivierobert, line 25: It would be better to make this happen in the form object. Prefer to keep controllers lean so with minimal business logic. Controllers should just do routing to the right form objects and return responses.  In contrast, form objects can be as complex as required and can make as many operations as needed. They are also far easier to test than controllers.

Tuubz:
Resolving conflicts... 🔨

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -4,10 +4,7 @@ module Api
   module V1
     class SearchResultsController < ApplicationController
       def create
-        upload_form = UploadForm.new(
-          search_engine: upload_form_params[:search_engine],
-          csv_file: upload_form_params[:csv_file]
-        )
+        upload_form = create_upload_form
```
olivierobert, line 7: This code is not idiomatic Ruby. A more idiomatic Ruby would be using the double pipe equal operator (read more about it: https://medium.com/@tali.scheer/ruby-or-equals-operator-6aed32af2a66):

```ruby

def create
  if upload_form.valid?
    #...
  end
end

private

def upload_form
  @upload_form ||= UploadForm.new(
    search_engine: upload_form_params[:search_engine],
    csv_file: upload_form_params[:csv_file]
  )
end 
```
 
Ruby has this motto: "Ask, Don’t Tell" (see https://www.cloudbees.com/blog/use-ask-dont-tell-policy-ruby), which aims to make code more elegant.

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class ClientServiceError < StandardError; end
+
+class SearchKeywordJob < ApplicationJob
+  queue_as :default
```
olivierobert, line 6: It is best to have a custom queue for scraping as other jobs (like sending out emails) could be on the default queue. Segregation by queus allows not block processes.

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def perform(search_result)
```
olivierobert, line 8: Since Redis serializes objects for background jobs a best practice is to only store the record ID instead of the whole object:

```suggestion
  def perform(search_result_id)
```

Then the job should fetch the correspond record from the database.

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class ClientServiceError < StandardError; end
```
olivierobert, line 3: Custom errors must be defined in `app/lib/project_name/error.rb`. See: https://github.com/nimblehq/mia-web/blob/development/app/lib/mia/errors.rb

This allows to have an overview of all possible application errors in a single place/

app/forms/upload_form.rb
```diff
@@ -45,7 +45,14 @@ def save
 
     search_results = keywords.map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
     SearchResult.import search_results
+    process_results(search_results)
```
olivierobert, line 48: You might be interested in using a more modern feature of Ruby to make the code more expressive:

```

def save
  content = csv_file.read
  keywords = CSV.parse(content).flatten.map(&:strip)

  keywords
    .map { |keyword| SearchResult.new(keyword: keyword, search_engine: search_engine) }
    .then(&method(:save_search_results)
	  .then(&method(:scrap_search_results)
end

private

def save_search_results(search_results)
   SearchResult.import(search_results)
end

def scrap_search_results(search_results)
  SearchKeywordJob.perform_later(search_result.id)
end
```

Here is a good article about using `tap` and `then` (available since Ruby 2.6): https://www.ananunesdasilva.com/posts/ruby-methods-tap-vs-then

app/models/search_result.rb
```diff
@@ -2,8 +2,11 @@
 
 class SearchResult < ApplicationRecord
   enum search_engine: { google: 'google', bing: 'bing' }
+  enum status: { in_progress: 'in_progress', completed: 'completed', failed: 'failed' }
 
   validates :keyword, presence: true
   validates :search_engine, presence: true
```
olivierobert, line 8: It is best to combine presence validators into one line when having 3 or more attributes using the same validation type:

```suggestion
  validates :keyword, :search_engine, :status, presence: true
```

olivierobert (changes requested):

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,19 @@
...
+  queue_as :scrapers
+
+  def perform(search_result_id)
+    search_result = SearchResult.find(search_result_id)
+    html_code = Search::SearchService.new(
+      keyword: search_result.keyword,
+      search_engine: search_result.search_engine
+    ).search
+
+    raise IcRailsHueyToby::Errors::ClientServiceError unless html_code
```
olivierobert, line 13: Do not raise exceptions for control flows. Exceptions are not "exceptional" cases 😉 they do not replace if/else blocks. Check this article: https://mensfeld.pl/2015/03/exceptions-should-not-be-expected-stop-using-them-for-control-flow-or-any-other-logic-handling-in-ruby/ and the linked articles at the bottom. 

```suggestion
    return mark_as_failed(search_result) unless html_code 
```

We could have a method `mark_as_failed` as it will be used also in the rescue block 💡

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    search_result.update html_code: html_code, status: :completed
+  rescue Faraday::ConnectionFailed, IcRailsHueyToby::Errors::ClientServiceError, NotImplementedError
```
olivierobert, line 16: If these are the expected errors, we could also refactor it using the `rescue_from` block: https://edgeguides.rubyonrails.org/active_job_basics.html#exceptions (I forgot about this)

```ruby
rescue_from IcRailsHueyToby::Errors::ClientServiceError do |_exception|
    mark_as_failed(search_result)
end
```

How about replacing `NotImplementedError` with `IcRailsHueyToby::Errors::ClientServiceError` as it could simplify the code? In addition, `Faraday::ConnectionFailed` is also replaced in #34

app/lib/ic_rails_huey_toby/errors.rb
```diff
@@ -0,0 +1,17 @@
...
+      attr_reader :original_error
+
+      def initialize(message = nil, original_error = nil)
+        super(message)
+
+        @original_error = original_error
+      end
+    end
+
+    class ClientServiceError < StandardError; end
```
olivierobert, line 15: It might be better to be more specific:

```suggestion
    class SearchServiceError < StandardError; end
```

olivierobert (changes requested):
Let's aim to have this PR merged before #34 as there are some dependencies between both

app/jobs/search_keyword_job.rb
Tuubz: I've simplified the exceptions to only use `IcRailsHueyToby::Errors::ClientServiceError` 🙏 

Is it really necessary to use `rescue_from`? If I understand correctly, it's defined outside of `def perform`

I'm a bit confused where it's supposed to get the `search_result` from:

```
  rescue_from IcRailsHueyToby::Errors::SearchServiceError do |_exception, search_result|
    mark_as_failed(search_result)
  end
```

olivierobert: It's not compulsory. It was just an idea as it is part of the DSL of ActiveJob.

olivierobert (approved):

hoangmirs (approved):

app/services/search/google_search_service.rb
```diff
@@ -13,9 +13,9 @@ def search
 
       response = Faraday.get(uri)
 
-      # TODO: Handle error in background job
-      # https://github.com/nimblehq/ic-rails-huey-toby/issues/19
       response.body if response.status == 200
+    rescue Faraday::ConnectionFailed
+      raise IcRailsHueyToby::Errors::SearchServiceError
```
hoangnguyen92dn, line 18: Just curious why don't we raise a more specific error message, i.e: "UnSupportedSearchEngine"? For security reason? 🤔

hoangnguyen92dn (approved):
The rest LGTM

app/services/search/google_search_service.rb
Tuubz: Originally I raised `Faraday::ConnectionFailed`, but this approach simplifies the code since the action we take is the same for all errors anyways (marking the search result as failed) 🙏 

**Reference**: https://github.com/nimblehq/ic-rails-huey-toby/pull/33#discussion_r1227471852

PR #34, opened 2023-06-07
[#28] [Backend] As a user, I can get the raw HTML of one Bing search page #28

### LINKED ISSUE #28: [Backend] As a user, I can get the raw HTML of one Bing search page
## Why

A search result must be made for one keyword in the CSV file, once the user uploads the file (Bing)

## Acceptance Criteria

- Make sure we're able to scrape the raw HTML for _one_ keyword (and store it in the `html_code` field):

## Design
N/A

## Resources

Tuubz:
- Close #28

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


config/application.yml
```diff
@@ -1,8 +1,8 @@
 default: &default
-  DB_NAME: "ic_rails_huey_toby"
+  DB_NAME: "postgres"
   DB_HOST: "localhost"
   DB_PORT: "5432"
-  DB_USERNAME: "postgres"
+  DB_USERNAME: "tuubz"
```
olivierobert, line 5: You need to change your local setup to use the default `postgres` username 🙈

Tuubz: Hahah yeah, I need to make work of it 🙈

app/services/search/bing_search_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def search
+      uri = URI.parse(@search_url)
+
+      response = Faraday.get(uri)
+
+      response.body if response.status == 200
+    rescue Faraday::ConnectionFailed
+      raise
```
olivierobert, line 18: The best practice is to raise a custom error. See https://github.com/nimblehq/mia-web/blob/development/app/lib/mia/errors.rb for reference.

Also it would be good to add the same logic to https://github.com/nimblehq/ic-rails-huey-toby/blob/develop/app/services/search/google_search_service.rb to ensure the codebase's consistency

olivierobert (commented):
The test suite is failing, so this must be checked before it can be merged: 

image

olivierobert (approved):

spec/services/search/search_service_spec.rb
```diff
@@ -11,6 +11,13 @@
       end
     end
 
+    context 'given the search engine is bing' do
+      it 'returns instance of BingSearchService' do
+        search_service = described_class.new(search_engine: 'bing', keyword: 'ruby')
+        expect(search_service).to be_a(Search::BingSearchService)
```
hoangmirs, line 17: Could you add a line break between Act and Assert parts? 🙏 The same for the above test

hoangmirs (approved):

hoangnguyen92dn (approved):

PR #37, opened 2023-06-15
[#15] [Backend] As a user, I can get scraped information from a Google search page

### LINKED ISSUE #15: [Backend] As a user, I can get scraped information from a Google search page

hoangnguyen92dn:
- Close #15 

## What happened 👀

- Add [ronin-web-user_agents](https://github.com/ronin-rb/ronin-web-user_agents) for generating User-Agent randomly
- Add [nokogiri](https://github.com/sparklemotion/nokogiri) to parse the result page. 
- Add new fields to the SearchResult model
- Add ParseService to scrape information from the search result page
- Add tests

## Insight 📝

Initially, I played around with [vessel](https://github.com/rubycdp/vessel) but the document is quite poor, and less support and answers from the community so decided back to Nokogiri.

Regarding information scraping from the search result page, the main idea is to collect elements from 3 containers with selectors:
- Top ads: `div.pla-unit-container a.pla-unit-title-link`
- Body ads: `div.data-text-ad div.v5yQqb`
- Search result: `div#search div.yuRUbf a[data-ved]`

Then extracting URLs by `href` selector.

## Proof Of Work 📹

- Top adwords count: 4
- Body adwords count: 0
- Non-adwords count: 9

screencapture-file-Users-hoangnguyen-Desktop-WebProjects-ic-rails-huey-toby-google-html-2023-06-19-17_23_42

Postman test

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/d5f063e7-81c3-4129-8ebb-e2950faeb7c1


app/services/search/google_search_service.rb
```diff
@@ -4,18 +4,99 @@ module Search
...
+
+      adwords_top_urls = parse_top_adwords(doc)
+      adwords_normal_urls = parse_normal_adwords(doc)
+      non_adwords_urls = parse_non_adwords(doc)
+
+      adwords_top_count = adwords_top_urls.size
+      adwords_normal_count = adwords_normal_urls.size
+      adwords_total_count = adwords_top_count + adwords_normal_count
+      non_adwords_count = non_adwords_urls.size
+      total_links_count = parse_all_links(doc)
```
hoangmirs, line 50: You can create a new function to return these attributes, then we can pass it to the SearchResult initializer

app/services/search/google_search_service.rb
```diff
@@ -4,18 +4,99 @@ module Search
...
+
     def initialize(keyword, language = 'en')
       @search_url = format(GOOGLE_SEARCH_URL, keyword: keyword, language: language)
     end
 
     def search
       uri = URI.parse(@search_url)
 
-      response = Faraday.get(uri)
+      response = Faraday.get(uri, nil, USER_AGENT_HEADER => USER_AGENT)
```
hoangmirs, line 27: We might need to random the user agent, as Google will detect spam when we use the same one.

output.html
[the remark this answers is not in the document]
Tuubz: It's for testing locally, right? 💭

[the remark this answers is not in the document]
hoangnguyen92dn: @hoangmirs Yeah. Since the search result is different so I cached this file for testing. I will remove then 🙏

app/services/search/google_search_service.rb
```diff
@@ -1,21 +1,103 @@
...
 module Search
   class GoogleSearchService
     GOOGLE_SEARCH_URL = 'https://www.google.com/search?q=%{keyword}&hl=%{language}&lr=%{language}'
 
+    USER_AGENT_HEADER = 'User-Agent'
+    USER_AGENT_TYPE = 'desktop_browser'
+
+    KEY_DIV_TOP_ADWORDS_CONTAINER = 'div.pla-unit-container'
+    KEY_DIV_NORMAL_ADWORDS_CONTAINER = 'div.v5yQqb'
+    KEY_DIV_NON_ADWORDS_CONTAINER = 'div.yuRUbf'
```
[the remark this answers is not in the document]
hoangnguyen92dn: @Tuubz Yeah. The current one doesn't work in all cases 😭 I will need to revise this parsing mechanism to get the information from body.

app/services/search/google_search_service.rb
```diff
@@ -1,21 +1,103 @@
...
+require 'user_agent_randomizer'
+
 module Search
   class GoogleSearchService
     GOOGLE_SEARCH_URL = 'https://www.google.com/search?q=%{keyword}&hl=%{language}&lr=%{language}'
 
+    USER_AGENT_HEADER = 'User-Agent'
+    USER_AGENT_TYPE = 'desktop_browser'
+
+    KEY_DIV_TOP_ADWORDS_CONTAINER = 'div.pla-unit-container'
```
olivierobert, line 12: In terms of naming, you might prefer the work `SELECTOR` as in DOM selector. Also it might be better to create an object containers all elements:

```ruby
SELECTORS = {
  adwords_top: 'div.pla-unit-container'
	adwords_body: 'div.v5yQqb'
}
```

It is similar to how selectors are structured in UI tests (see https://github.com/nimblehq/mia-web/blob/development/spec/systems/mia_campaign/reject_campaign/reject_campaign_spec.rb#L4)

app/services/search/google_search_service.rb
hoangnguyen92dn: @hoangmirs I noticed that the search page is returning different results since we are generating UA randomly for usage. This is the challenge to achieve the parsing logic to make it works for all cases. Do you have any advice to tackle this? 🙏

output.html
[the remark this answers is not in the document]
hoangnguyen92dn: Nevermind 😄 I changed the UA generator to generate specific UA for chrome browser so it works

app/services/search/google_search_service.rb
hoangnguyen92dn: Nevermind 😄 I changed the UA generator to generate specific UA for chrome browser so it works

app/jobs/search_keyword_job.rb
```diff
@@ -5,22 +5,36 @@ class SearchKeywordJob < ApplicationJob
...
+      html_code: html_code
+    ).parse
+  end
+
+  def mark_as_completed(search_result, html_code, result)
+    search_result.update html_code: html_code, adwords_top_urls: result[:adwords_top_urls],
+                         adwords_top_count: result[:adwords_top_count], adwords_total_count: result[:adwords_total_count],
+                         non_adwords_urls: result[:non_adwords_urls], non_adwords_count: result[:non_adwords_count],
+                         total_links_count: result[:total_links_count], status: :completed
   end
```
hoangmirs, line 38: As you are listing all the keys of `result`, should we just pass the whole `result` instead?
```suggestion
  def mark_as_completed(search_result, html_code, result)
    search_result.update(result.merge(html_code: html_code, status: :completed))
  end
```

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,74 @@
...
+    SELECTORS = {
+      adwords_top: 'div.pla-unit-container a.pla-unit-title-link',
+      adwords_body: 'div.data-text-ad div.v5yQqb',
+      non_adwords: 'div#search div.yuRUbf a[data-ved]',
+      href: 'href'
+    }.freeze
+
+    def initialize(html_code)
+      @html_code = html_code
+      raise IcRailsHueyToby::Errors::ParseServiceError unless html_code
```
hoangmirs, line 14: Do we need this error? As we check the nil `html_code`, can we return nil in this case?
IMO, this error should be raised when there is something wrong with the parser 🤔

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      adwords_top_urls = parse_top_adwords(doc)
+      adwords_body_urls = parse_body_adwords(doc)
+      non_adwords_urls = parse_non_adwords(doc)
+
+      adwords_top_count = adwords_top_urls.count
+      adwords_body_count = adwords_body_urls.count
+      adwords_total_count = adwords_top_count + adwords_body_count
+      non_adwords_count = non_adwords_urls.count
+      total_links_count = adwords_total_count + non_adwords_count
```
hoangmirs, line 37: Should we split these blocks into 2 methods so that we don't need to add `rubocop:disable Metrics/MethodLength`?

app/jobs/search_keyword_job.rb
```diff
@@ -5,22 +5,36 @@ class SearchKeywordJob < ApplicationJob
...
+  end
+
+  def parse(search_engine, html_code)
+    Parse::ParseService.new(
+      search_engine: search_engine,
+      html_code: html_code
+    ).parse
+  end
+
+  def mark_as_completed(search_result, html_code, result)
```
olivierobert, line 33: This method has an envy issue and does not do what its name says. It is best to leave as it was before (so just updating the stattus) but have another method to add the search result to the `search_result` record, e.g., `set_parsing_result(search_result, html_code, result)`

app/services/parse/google_parse_service.rb
olivierobert: If we want to add a check (which is not compulsory), we could rely on Nokogiri to tell us if there is any error (see https://nokogiri.org/tutorials/ensuring_well_formed_markup.html#gentle-well-formedness)

app/services/parse/google_parse_service.rb
olivierobert: We can definitely and should break into parsing adword related and non-adwords content. The total links parsing can also be another method.

spec/jobs/search_keyword_job_spec.rb
```diff
@@ -8,14 +8,77 @@
...
         expect(search_result.status).to eq('in_progress')
 
         expect { described_class.perform_now(search_result.id) }
           .to change { search_result.reload.status }
           .to('completed').and change { search_result.html_code }
-          .to(include('<title>ruby - Google Search</title>'))
+          .to(include('<title>tivi - Google Search</title>'))
+      end
+
+      it 'updates the adwords_top_urls correctly', vcr: 'services/search/google/valid' do
```
olivierobert, line 21: We can omit the adverb `correctly` 😅 

```suggestion
      it 'updates the adwords_top_urls with the parsed content from the seach page', vcr: 'services/search/google/valid' do
```

olivierobert (changes requested):

.rubocop.yml
```diff
@@ -46,7 +46,7 @@ RSpec/ContextWording:
     - given
 
 RSpec/ExampleLength:
-  Max: 7
+  Enabled: false
```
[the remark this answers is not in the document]
hoangnguyen92dn: @hoangmirs 🆗 So let's me increase the max `ExampleLength` then 😄

app/services/parse/google_parse_service.rb
hoangnguyen92dn: @olivierobert By relying on Nokogiri to validate the `html_code`, even with the valid one, it's raising issues such as `Tag svg invalid` if the html_code contains svg tag and so on; when passing `nil`, it will skip the validations and no errors will be raised. 💭 It's so confusing if we rely on well-formed validations to raise the error if any.

.rubocop.yml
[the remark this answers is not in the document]
hoangnguyen92dn: Fixed [1adda90](https://github.com/nimblehq/ic-rails-huey-toby/pull/37/commits/1adda9051fe30fc4a4a4e39385890f201f64918d)

app/jobs/search_keyword_job.rb
hoangnguyen92dn: Fixed [7c89c40](https://github.com/nimblehq/ic-rails-huey-toby/pull/37/commits/7c89c401ec959f8b86dbe55a5dbda7e802f15af0)

app/services/parse/google_parse_service.rb
hoangnguyen92dn: Fixed [c8e81d8](https://github.com/nimblehq/ic-rails-huey-toby/pull/37/commits/c8e81d88c67f79a754298d61aa528cd44d75cec9)

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+    }.freeze
+
+    def initialize(html_code)
+      @html_code = html_code
+      raise IcRailsHueyToby::Errors::ParseServiceError unless html_code
+    end
+
+    def parse
+      parse_html
+    end
```
Tuubz, line 19: Optional: Remove this, and rename `parse_html` to `parse` and make the function public

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      top_ads_container = doc.css(SELECTORS[:adwords_top])
+
+      parse_url(top_ads_container)
+    end
+
+    def parse_body_adwords(doc)
+      body_ads_container = doc.css(SELECTORS[:adwords_body])
+
+      parse_url(body_ads_container)
+    end
```
Tuubz, line 45: Should these be placed below `parse_adwords`? 💭

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+        non_adwords_urls: non_adwords_urls,
+        non_adwords_count: non_adwords_count
+      }
+    end
+
+    def parse_total_links_count(adwords_result, non_adwords_result)
+      adwords_result[:adwords_total_count] + non_adwords_result[:non_adwords_count]
+    end
+
+    def parse_url(container)
```
Tuubz, line 78: Would this make more sense? 🙏 
```suggestion
    def parse_urls(container)
```

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      non_adwords_count = non_adwords_urls.count
+
+      {
+        non_adwords_urls: non_adwords_urls,
+        non_adwords_count: non_adwords_count
+      }
+    end
+
+    def parse_total_links_count(adwords_result, non_adwords_result)
+      adwords_result[:adwords_total_count] + non_adwords_result[:non_adwords_count]
```
Tuubz, line 75: Could we do this? So it's aligned with `parse_non_adwords` & `parse_adwords`

```suggestion
      total_links_count = adwords_result[:adwords_total_count] + non_adwords_result[:non_adwords_count]

      { total_links_count: total_links_count }
```

app/jobs/search_keyword_job.rb
```diff
@@ -5,22 +5,38 @@ class SearchKeywordJob < ApplicationJob
...
   def perform(search_result_id)
     search_result = SearchResult.find(search_result_id)
-    html_code = Search::SearchService.new(
-      keyword: search_result.keyword,
-      search_engine: search_result.search_engine
-    ).search
+    html_code = search(search_result.keyword, search_result.search_engine)
 
-    return mark_as_failed(search_result) unless html_code
+    result = parse(search_result.search_engine, html_code)
```
Tuubz, line 10: Could we rename this to `parsing_result` / `parse_result`, for more clarity? 🙇 

Same goes for the parameter inside `set_parsing_result` 🙏

app/jobs/search_keyword_job.rb
```diff
@@ -5,22 +5,38 @@ class SearchKeywordJob < ApplicationJob
...
+
+  def parse(search_engine, html_code)
+    Parse::ParseService.new(
+      search_engine: search_engine,
+      html_code: html_code
+    ).parse
+  end
+
+  def set_parsing_result(search_result, html_code, result)
+    search_result.update result.merge({ html_code: html_code })
```
Tuubz, line 34: We could potentially do `.merge({ html_code: html_code })` inside `parse_service`, since we pass it as a parameter already anyways, thoughts? 💭

app/services/search/google_search_service.rb
```diff
@@ -4,18 +4,26 @@ module Search
...
     end
 
     def search
       uri = URI.parse(@search_url)
 
-      response = Faraday.get(uri)
+      response = Faraday.get(uri, nil, USER_AGENT_HEADER => generate_user_agent)
 
-      response.body if response.status == 200
+      @html_code = response.body if response.status == 200
```
Tuubz, line 18: Is this change necessary? If I'm not mistaken we can omit `@html_code`

Tuubz (commented):
@hoangnguyen92dn Small renaming/refactor suggestions, I'll do some testing now 🚀

app/services/parse/google_parse_service.rb
olivierobert: Understood. So let's raise an exceptopn then but I recommend to place it before local variales assignments and add a line break for readability:

```ruby
def initialize(html_code)
  raise IcRailsHueyToby::Errors::ParseServiceError unless html_code

  @html_code = html_code
end
```

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      adwords_result.merge(non_adwords_result, { total_links_count: total_links_count })
+    end
+
+    def parse_top_adwords(doc)
+      top_ads_container = doc.css(SELECTORS[:adwords_top])
+
+      parse_url(top_ads_container)
+    end
+
+    def parse_body_adwords(doc)
```
olivierobert, line 41: Always avoid plural as it is often the root cause of typos: https://nimblehq.co/compass/development/code-conventions/#convention-15

```suggestion
    def parse_body_adword(doc)
```

or

```suggestion
    def parse_body_adword_list(doc)
```

app/services/parse/google_parse_service.rb
olivierobert: Ditto from above on avoiding plurals:

```suggestion
    def parse_url_list(container)
```

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      }
+    end
+
+    def parse_total_links_count(adwords_result, non_adwords_result)
+      adwords_result[:adwords_total_count] + non_adwords_result[:non_adwords_count]
+    end
+
+    def parse_url(container)
+      container.map do |element|
+        element[SELECTORS[:href]]
```
olivierobert, line 80: We might prefer a one-liner for readability:

```ruby
container.map { |element| element[SELECTORS[:href]] }
```

app/services/search/google_search_service.rb
hoangnguyen92dn: @Tuubz Good catch 🙈

app/jobs/search_keyword_job.rb
hoangnguyen92dn: @Tuubz By merging the `html_code` inside google_parse_service, we will need to update the test case to expect the html_code from returning result. Because the html_code is quite long so we should skip that testcase. What do you think?

Tuubz (approved):
🚀

app/jobs/search_keyword_job.rb
```diff
@@ -5,22 +5,38 @@ class SearchKeywordJob < ApplicationJob
...
+
+  def parse(search_engine, html_code)
+    Parse::ParseService.new(
+      search_engine: search_engine,
+      html_code: html_code
+    ).parse
+  end
+
+  def set_parse_result(search_result, html_code, parse_result)
+    search_result.update parse_result.merge({ html_code: html_code })
```
hoangmirs, line 34: Let's add curly brackets here for readable 🙏 
```suggestion
    search_result.update(parse_result.merge({ html_code: html_code }))
```

hoangmirs (approved):

app/jobs/search_keyword_job.rb
```diff
@@ -5,25 +5,41 @@ class SearchKeywordJob < ApplicationJob
...
+  def parse(search_engine, html_code)
+    Parse::ParseService.new(
+      search_engine: search_engine,
+      html_code: html_code
+    ).parse
+  end
+
+  def set_parse_result(search_result, html_code, parse_result)
+    search_result.update(parse_result.merge({ html_code: html_code }))
+    mark_as_completed(search_result)
```
olivierobert, line 35: This method should not be here 😅 Based on the method name, it is a very unexpected side effect. It would be best to have at line 12:

```ruby
set_parse_result(search_result, html_code, parse_result)
mark_as_completed(search_result)
```

app/services/parse/google_parse_service.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def parse
+      doc = Nokogiri::HTML(html_code)
+
+      adwords_result = parse_adword_list(doc)
+      non_adwords_result = parse_non_adword_list(doc)
+      total_links_count = parse_total_links_count(adwords_result, non_adwords_result)
+
+      adwords_result.merge(non_adwords_result, total_links_count)
```
olivierobert, line 25: It is a not odd at first as I was looking where was the first hash 😅 How about:

```ruby
Hash.new.merge(adwords_result, non_adwords_result, total_links_count)
```

olivierobert (commented):
Just some minor improvement suggestions. Let me know what you think then I will approve it.

app/services/parse/google_parse_service.rb
hoangnguyen92dn: Oh. It looks clearer that we gonna merge 3 hashes.

app/jobs/search_keyword_job.rb
hoangnguyen92dn: Hmmm...Initially, I thought updating the parse result means that the parsing process has been done and we could update the status as completed 💭 Indeed we should separate the updating state to avoid confusion from the method name. Let me update then 👍

hoangnguyen92dn: Fixed [1c1bdbc](https://github.com/nimblehq/ic-rails-huey-toby/pull/37/commits/1c1bdbc3f1dbdf7ee07606b218b9683f5ea9bf39)

app/services/parse/google_parse_service.rb
hoangnguyen92dn: Fixed [b5004c9](https://github.com/nimblehq/ic-rails-huey-toby/pull/37/commits/b5004c9c50c78d795cdabff28210dd26404e18b7)

olivierobert (approved):

PR #39, opened 2023-06-22
[#23] [Backend] As a user, I can get scraped information from a Bing search page

### LINKED ISSUE #23: [Backend] As a user, I can get scraped information from a Bing search page
## Why

Scrape the details from the search result for every keyword in the CSV file (from Bing)

## Acceptance Criteria

- Scrape and store the following details as a result, from the raw HTML for every keyword:
    - Number of AdWords advertisers in the top position.
    - Total number of AdWords advertisers on the page.
    - URLs of the AdWords advertisers in the top position. (array to string)
    - Number of the non-AdWords results on the page.
    - URLs of the non-AdWords results on the page. (array to string)
    - Total number of links (all of them) on the page 
    - HTML code of the page/cache of the page. (#14)
    - Search engine used (Google or Bing)

**Note**:  Use `nokogiri` to scrape the raw HTML

## Design
N/A

## Resources

Image

Tuubz:
Close #23 

## What happened 👀

- Add parsing service for Bing, very similar to #37, but the selectors are different:
    - `adwords_top` are the ads that are shown on the side
    - `adwords_body` are the ads that are mixed between the normal search results 
    - `non_adwords` are the normal search results and the recommended for you results

- Create `UserAgentGenerator`, to remove redundancy.

## Insight 📝
Refactor parameters of search service to be named, because there was a bug in the test:

Instead of `<title>{:keyword=&gt;&quot;watch&quot;} - Search</title>` it must be `'<title>watch - Search</title>'`

## Proof Of Work 📹

Screen Shot 2023-06-22 at 14 37 50


.rubocop.yml
```diff
@@ -46,7 +46,7 @@ RSpec/ContextWording:
     - given
 
 RSpec/ExampleLength:
-  Max: 30
+  Max: 37
```
olivierobert, line 49: It stop making sense if you keep increasing this. I recommend revisit the test by breaking them down.

app/services/search/user_agent_generator.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Search
+  module UserAgentGenerator
```
olivierobert, line 4: This extraction is not idiomatic Ruby 😅 FFor concerns, the naming is not Ruby-ish enough. Such modules are usually named with the prefix `has` (e.g., `hasSettings`) or the suffix `able` (e.g., `include Disardable`). You can read mode about it in this article: https://www.akshaykhot.com/how-rails-concerns-work-and-how-to-use-them/.

We can also make the argument that this looks more like a case for a helper or PORO:

```ruby
class UserAgentHeaderGenerator
    USER_AGENT_HEADER = 'User-Agent'

    def initialize
			@header_name = USER_AGENT_HEADER
    end

		def call
				{ @header_name => Ronin::Web::UserAgents.chrome.random }
    end
end
```

Then in the service classes:

```ruby
headers = {}
user_agent_header = UserAgentHeaderGenerator.call()

response = Faraday.get(uri, nil, headers.merge(user_agent_header))
```

> **Note**
> When a class uses the method `call`, we can use this syntax `UserAgentHeaderGenerator.()`. This is any FYI, not recommended though 😅

spec/services/parse/bing_parse_service_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      it 'returns parse result', vcr: 'services/search/bing/valid' do
+        search_service = Search::BingSearchService.new(keyword: 'watch')
+        html_code = search_service.search
+
+        parse_service = described_class.new(html_code)
+        parse_result = parse_service.parse
+
+        expect(parse_result).to eq(
+          {
+            adwords_top_urls: [
```
olivierobert, line 17: This g=hurts the test readability. Can you edit the VCR recording to simplify this data?

app/services/parse/bing_parse_service.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def parse
+      doc = Nokogiri::HTML(html_code)
+
+      adword_list = parse_adword_list(doc)
+      non_adword_list = parse_non_adword_list(doc)
+      total_links_count = parse_total_links_count(adword_list, non_adword_list)
+
+      Hash.new.merge(adword_list, non_adword_list, total_links_count) # rubocop:disable Style/EmptyLiteral
```
hoangmirs, line 25: Can we change to `{}` instead of `Hash.new` to avoid [this warning](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyLiteral)?
```suggestion
      {}.merge(adword_list, non_adword_list, total_links_count)
```

spec/services/parse/bing_parse_service_spec.rb
Tuubz: @olivierobert I could be wrong, but it's not possible to edit the VCR recording, because the body is encoded:

Screen Shot 2023-06-23 at 10 52 01

However, I can re-fator the URLS in the a constant, what do you think? 🙏

Tuubz: Or... I guess I could try to decode it, edit it, and encode it again 💭

app/services/search/google_search_service.rb
```diff
@@ -4,26 +4,19 @@ module Search
...
-    def initialize(keyword, language = 'en')
+    def initialize(keyword:, language: 'en')
       @search_url = format(GOOGLE_SEARCH_URL, keyword: keyword, language: language)
     end
 
     def search
       uri = URI.parse(@search_url)
 
-      response = Faraday.get(uri, nil, USER_AGENT_HEADER => generate_user_agent)
+      headers = {}.merge(UserAgentHeaderGenerator.call)
```
hoangnguyen92dn, line 14: Since we only have user-agent in the header, should we get rid of `{}.merge`? 

```suggestion
headers = UserAgentHeaderGenerator.call
```

app/services/search/bing_search_service.rb
```diff
@@ -4,14 +4,15 @@ module Search
...
-    def initialize(keyword, language = 'en')
+    def initialize(keyword:, language: 'en')
       @search_url = format(BING_SEARCH_URL, keyword: keyword, language: language)
     end
 
     def search
       uri = URI.parse(@search_url)
 
-      response = Faraday.get(uri)
+      headers = {}.merge(UserAgentHeaderGenerator.call)
```
hoangnguyen92dn, line 14: Same here 😄 

Btw, the error should be

```
raise IcRailsHueyToby::Errors::SearchServiceError
```

instead of

```
raise IcRailsHueyToby::Errors::ClientServiceError
```

💭

hoangnguyen92dn (commented):
The rest LGTM

hoangnguyen92dn (approved):

olivierobert (approved):

hoangmirs (approved):

PR #40, opened 2023-06-22
[#26][#29] [Backend] As a user, I can get the raw HTML of multiple Google & Bing search pages

### LINKED ISSUE #26: [Backend] As a user, I can get the raw HTML of multiple Google search pages
## Why

A search result must be made for every keyword in the CSV file, once the user uploads the file (Google).

## Acceptance Criteria

- Make sure we're able to scrape the raw HTML for _multiple_ keywords (and store it in the `html_code` field):

⚠ You must find a way to work around the limitations of mass-searching keywords as Google prevents it.

## Design
N/A

## Resources

### LINKED ISSUE #29: [Backend] As a user, I can get the raw HTML of multiple Bing search pages
## Why

A search result must be made for every keyword in the CSV file, once the user uploads the file (Bing).

## Acceptance Criteria

- Make sure we're able to scrape the raw HTML for _multiple_ keywords (and store it in the `html_code` field):

⚠ You must find a way to work around the limitations of mass-searching keywords as Bing prevents it.

## Design
N/A

## Resources

hoangnguyen92dn:
- Close #26 
- Close #29 

## What happened 👀

Make sure we're able to scrape the raw HTML for multiple keywords (and store it in the html_code field).

## Insight 📝

Since Google blocked massive searches in a short time from a client, there are several approaches to bypass the search blocking:

- Add the delay between jobs: Random number in seconds in range 1 second to 10 minutes (1..600) ✅ 

```
    results.map(&:id).each do |id|
      delay_in_seconds = rand(1..600).seconds # Maximum delay is 10 minutes
      SearchKeywordJob.set(wait: delay_in_seconds).perform_later(id)
    end
```
- Add UserAgent generated randomly ✅ 

```
    def generate_user_agent
      Ronin::Web::UserAgents.chrome.random
    end
```

- ~~Add the X-forwarded-for with fake IP addresses into the header~~ ✅ 
  - Since adding delay between jobs is sufficient so we could get rid of this approach

- Proxy: Free proxy server changes frequently so can’t use this 🚫 

## Proof Of Work 📹

**GOOGLE**
Total of keywords: 295
Success results: 170

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/d53eee8a-f5b0-495b-98d8-623e7d016172

**BING**
Total of keywords: 300
Success results: All, but some of them had to retry

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/bd1ed041-a4d1-425f-aa5b-d9d313f03753

### DIFF
--- a/app/forms/upload_form.rb
+++ b/app/forms/upload_form.rb
@@ -40,8 +40,9 @@ class UploadForm
   end
 
   def scrape_search_results(results)
-    results.map(&:id).each do |id|
-      SearchKeywordJob.perform_later(id)
+    results.each_with_index do |result, index|
+      delay_in_seconds = 3.seconds * index
+      SearchKeywordJob.set(wait: delay_in_seconds).perform_later(result.id)
     end
   end
 
--- a/app/forms/upload_form.rb
+++ b/app/forms/upload_form.rb
@@ -40,9 +40,9 @@ class UploadForm
   end
 
   def scrape_search_results(results)
-    results.each_with_index do |result, index|
-      delay_in_seconds = 3.seconds * index
-      SearchKeywordJob.set(wait: delay_in_seconds).perform_later(result.id)
+    results.map(&:id).each do |id|
+      delay_in_seconds = rand(1..600).seconds # Maximum delay is 10 minutes
+      SearchKeywordJob.set(wait: delay_in_seconds).perform_later(id)
     end
   end
 
--- a/app/forms/upload_form.rb
+++ b/app/forms/upload_form.rb
@@ -41,9 +41,12 @@ class UploadForm
 
   def scrape_search_results(results)
     results.map(&:id).each do |id|
-      delay_in_seconds = rand(1..600).seconds # Maximum delay is 10 minutes
-      SearchKeywordJob.set(wait: delay_in_seconds).perform_later(id)
+      SearchKeywordJob.set(wait: generate_delay_in_seconds).perform_later(id)
     end
+  end
+
+  def generate_delay_in_seconds
+    rand(1..600).seconds # Maximum delay is 10 minutes
   end
 
   def csv_file_content_type
--- a/spec/forms/upload_form_spec.rb
+++ b/spec/forms/upload_form_spec.rb
@@ -90,6 +90,17 @@ RSpec.describe UploadForm, type: :model do
       end
     end
 
+    context 'given valid attributes with massive list of keywords' do
+      it 'runs search keyword jobs with random delay' do
+        upload_form = described_class.new(
+          search_engine: 'google',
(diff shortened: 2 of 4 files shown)


app/forms/upload_form.rb
```diff
@@ -40,8 +40,9 @@ def save_search_results(results)
   end
 
   def scrape_search_results(results)
-    results.map(&:id).each do |id|
-      SearchKeywordJob.perform_later(id)
+    results.each_with_index do |result, index|
+      delay_in_seconds = 3.seconds * index
```
olivierobert, line 44: I recommend using some randomization too

spec/forms/upload_form_spec.rb
```diff
@@ -90,6 +90,20 @@
...
     end
 
+    context 'given valid attributes with massive list of keywords' do
+      it 'runs search keyword jobs with random delay' do
+        upload_form = described_class.new(
+          search_engine: 'google',
+          csv_file: fixture_file_upload('upload_valid_massive_keywords.csv', 'text/csv')
+        )
+
+        allow_any_instance_of(described_class).to receive(:generate_delay_in_seconds).and_return(5) # rubocop: disable RSpec/AnyInstance
```
Tuubz, line 100: 🚀 
```suggestion
        allow(upload_form).to receive(:generate_delay_in_seconds).and_return(5)

```

Tuubz (approved):
I left one comment, rest LGTM 🚀

spec/forms/upload_form_spec.rb
olivierobert: `allow_any_instance_of` is a massive code smell., never use it unless there is no other way 🙈

olivierobert (approved):

app/forms/upload_form.rb
```diff
@@ -41,10 +41,14 @@ def save_search_results(results)
 
   def scrape_search_results(results)
     results.map(&:id).each do |id|
-      SearchKeywordJob.perform_later(id)
+      SearchKeywordJob.set(wait: generate_delay_in_seconds).perform_later(id)
```
hoangmirs, line 44: 🤔 With this, jobs can be in the queue simultaneously, as `generate_delay_in_seconds` return a random second, and the random number can be roundly equal, right?
I think a delay between jobs could be better. But let's go with this as we want to find a way to achieve it 💪

hoangmirs (approved):

app/forms/upload_form.rb
hoangnguyen92dn: @hoangmirs Indeed. The generated number can be even the same values. With current success rate, it's acceptable so I think we could go with the current implementation 🚀

Issue #11, opened 2023-05-25
[Backend] As a user, I can sign-in with Google


Tuubz:
## Why

Allow a user to sign in to their account (with Google), once authenticated they can interact with the application.

## Acceptance Criteria

- Check if the user already exists, if not return an error.

- Re-use the same endpoint `POST api/v1/users/auth/google_oauth2/callback`:

   - **Request**: Google client ID & Google client secret
  - **Response**: See example below:

```
{
  "access_token": "access token",
  "token_type": "bearer",
  "expires_in": 7200,
  "refresh_token": "refresh token",
  "scope": "scopes"
}
```

**Note**: Use `doorkeeper` to generate the response (based on the user ID)

- Write unit and integration tests

## Design
N/A

## Resources
- https://github.com/zquestz/omniauth-google-oauth2
- https://www.youtube.com/watch?v=CnZnwV38cjo
- https://blog.postman.com/how-to-access-google-apis-using-oauth-in-postman/

PR #43, opened 2023-06-27
[#9] [Backend] As a user, I can sign-up with Google

### LINKED ISSUE #9: [Backend] As a user, I can sign-up with Google
## Why

Allow a user to register a new account (with Google), once authenticated they can interact with the application.

## Acceptance Criteria

- Set up Google API credentials: https://console.developers.google.com/, for the client ID & client secret

- Add a new field to the table for users, to differentiate between normal and Google accounts. (ex : `type`)

    → Optional: Since Google will not provide the password, we can generate a token for it.

- Create a new endpoint: `POST api/v1/users/auth/google_oauth2/callback`:

   - **Request**: Google client ID & Google client secret

```
{
  "token_type": "bearer",
  "expires_in": 7200,
  "scope": "scopes"
}
```

- Restricted to email addresses with the domain `@nimblehq.co`, otherwise show an error message

## Design
N/A

## Resources
- https://github.com/zquestz/omniauth-google-oauth2
- https://www.youtube.com/watch?v=CnZnwV38cjo
- https://blog.postman.com/how-to-access-google-apis-using-oauth-in-postman/

hoangnguyen92dn:
- Close #9 

## What happened 👀

- Add [omniauth](https://github.com/omniauth/omniauth) and  [omniauth-google-oauth2](https://github.com/zquestz/omniauth-google-oauth2) gem to support Google OAuth2
- Since we are building an API application so there is no UI for the Google sign-up/sign-in buttons, to handle CTA to open the user's consent popup, [Postman supports OAuth2 ](https://blog.postman.com/how-to-access-google-apis-using-oauth-in-postman/) to retrieve authorization code to exchange the token for accessing Google APIs, the application will need to access user's profile and email to validate users. To understand Google OAuth2 flow, please visit [OAuth2 playground ](https://developers.google.com/oauthplayground/) page.
- New project created on Google Cloud Console to config redirect URIs

image

## Insight 📝

- In the [Google authorization sequence](https://developers.google.com/identity/protocols/oauth2), `omniauth` will handle the steps exchange token then access Google APIs to retrieve data via `users/auth/google_oauth2/callback`, then we need to handle the user's profile in the `OmniauthCallbacksController`.

- Omniauth adding `state` validation to avoid CSRF issue (https://github.com/omniauth/omniauth-oauth2/blob/3a43234ab5dd36a75f9c125c58fcfe1a37b26805/lib/omniauth/strategies/oauth2.rb#L87), so when processing the user's consent step via Postman it will throw the `csrf_detected`; to bypass this, config `provider_ignores_state: true` in `devise.rb` to allow the application receive the callback request.

- In order to use Postman, we need to add these configs

image
image

*Notice:* `Doorkeeper` will be added on #11 to returns JWT token for user access

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/a2b37754-b791-4af5-9ddd-3951ea7195e0


app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        end
+
+        private
+
+        def auth
+          @auth ||= request.env['omniauth.auth']
+        end
+
+        def render_success
+          render json: { success: true }
```
Tuubz, line 26: Just to be sure we're aligned, we will return a token in the next issue, right? 🙏 

#11

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module Api
+  module V1
+    module Users
+      class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+        respond_to :json
+        def google_oauth2
+          @user = User.from_omniauth(auth)
+
+          # TODO: Return JWT token on both success and error
```
Tuubz, line 11: What's the reason we will return the token in error? 🙏

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
Tuubz, line 26: Is there a suitable `status` we can apply here ? 💭

spec/fabricators/user_fabricator.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+Fabricator(:user) do
+  id { 1 }
+  provider { 'google_oauth2' }
```
Tuubz, line 5: Could we keep it with `google`? 🙏

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
olivierobert, line 11: JWT Tokens and OAuth tokens are not the same thing ⚠️ So I recommend to to do more research to understand the differences to deepen your understanding of API. Here is quick insight: https://stackoverflow.com/a/39914013

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+          if @user.persisted?
+            render_success
```
olivierobert, line 13: Why extracting to methods when it is a single line of code 😅

app/models/user.rb
```diff
@@ -0,0 +1,28 @@
...
+class User < ApplicationRecord
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :validatable,
+         :omniauthable, omniauth_providers: [:google_oauth2]
+
+  def self.from_omniauth(auth)
+    find_existing_user(auth) || create_new_user(auth)
+  end
+
+  private_class_method def self.find_existing_user(auth)
```
olivierobert, line 12: The team never use `private_class_method` and I have not seen it used in the Ruby community either 😅 Why use that syntax?

app/models/user.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class User < ApplicationRecord
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :validatable,
+         :omniauthable, omniauth_providers: [:google_oauth2]
+
+  def self.from_omniauth(auth)
+    find_existing_user(auth) || create_new_user(auth)
```
olivierobert, line 9: ActiveRecord has a method `find_or_initialize_by` that can remove the OR block 💡

db/migrate/20230626042619_devise_create_users.rb
```diff
@@ -0,0 +1,48 @@
...
+
+class DeviseCreateUsers < ActiveRecord::Migration[7.0]
+  def change
+    create_table :users do |t|
+      ## Database authenticatable
+      t.string :email,              null: false, default: ""
+      t.string :encrypted_password, null: false, default: ""
+      t.string :full_name
+      t.string :uid
+      t.string :avatar_url
```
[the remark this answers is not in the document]
olivierobert: Let's keep them still, but mark them as nullable. We can imagine that for users who login with email + password, there could be an endpoint for them to update their user profile with this data.

spec/models/user_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      'family_name' => 'Smith',
+      'profile' => 'https://plus.google.com/+JohnSmith',
+      'picture' => 'https://lh4.googleusercontent.com/photo.jpg?sz=50',
+      'email' => 'john@example.com',
+      'email_verified' => 'true',
+      'locale' => 'en',
+      'hd' => 'company.com'
+    }
+  }
+}.freeze
```
olivierobert, line 51: It would be best to keep these hashes into https://github.com/nimblehq/ic-rails-huey-toby/tree/develop/spec/fixtures as files

spec/fabricators/user_fabricator.rb
```diff
@@ -0,0 +1,11 @@
...
+  uid { '100000000000000000000' }
+  email { 'john@example.com' }
```
olivierobert, line 7: This will break future tests when we need to create multiple users 💣 all attributes must be randomized

spec/models/user_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+RSpec.describe User, type: :model do
+  describe '.from_omniauth' do
+    context 'given an existing user is found' do
+      it 'returns the existing user' do
+        auth = OmniAuth::AuthHash.new(OMNIAUTH_HASH)
+
+        existing_user = Fabricate(:user)
```
olivierobert, line 59: The specific attribute for names, email, etc. must be passed to make the test more understandable and to override the randomized name attributes from the fixture definition:

```suggestion
        existing_user = Fabricate(:user, email: john@example.com, ...)
```

spec/models/user_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        expect(described_class.from_omniauth(auth)).to eq(existing_user)
+      end
+    end
+
+    context 'given a new user is created' do
```
olivierobert, line 64: How about instead:

```suggestion
    context 'given a new user' do
```

spec/models/user_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      it 'returns a new user with the provided attributes' do
```
olivierobert, line 65: How about:

```suggestion
      it 'creates a new user with the provided attributes' do
```

olivierobert (changes requested):

spec/models/user_spec.rb
hoangnguyen92dn: @olivierobert I tried to create a file containing the hash, unfortunately, when reading the content file and converting to hash, it requires using `eval` and this will cause of security risk warning.

```ruby
        fixture = file_fixture('omniauth_hash.txt').read
        auth = OmniAuth::AuthHash.new(eval(fixture))
```

We could convert to yaml format but I'd prefer to keep the hash as a reference from the [ominauth auth-hash example.](https://github.com/zquestz/omniauth-google-oauth2#auth-hash). So I think we can leave it as it is for now 🙏

app/models/user.rb
hoangnguyen92dn: @olivierobert Since we have `from_omniauth()` as the self methods so that's why I added `private_class_method` to make the `find_existing_user` and `create_new_user` can be invoked 😅 I updated to use `class << self` with `private` section instead 🙏

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
hoangnguyen92dn: I was supposed to add more lines of code with generating tokens, so I extracted new methods for future updates 😄

app/models/user.rb
hoangnguyen92dn: @olivierobert Since we will need to create a new record and save it to db if there is no existing user, I will update to use `find_or_create_by` instead ([find_or_initialize_by](https://guides.rubyonrails.org/active_record_querying.html#find-or-initialize-by) just create new record in memory) 😄

spec/models/user_spec.rb
hoangnguyen92dn: P/s: Otherwise, we could use randomzied auth_hash from [Faker](Faker::Omniauth.google )

spec/fabricators/user_fabricator.rb
hoangnguyen92dn: Similarly to https://github.com/nimblehq/ic-rails-huey-toby/pull/43#discussion_r1247512922, all attributes will be randomized by using Faker

spec/fabricators/user_fabricator.rb
hoangnguyen92dn: @Tuubz It's the provider is being used to determine google OAuth2 from omniauth:
- https://github.com/zquestz/omniauth-google-oauth2#auth-hash

I'd prefer to keep it for consistency and easier to generate test data with Faker:

- https://github.com/faker-ruby/faker/blob/main/doc/default/omniauth.md

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
hoangnguyen92dn: @Tuubz Ideally, when receiving a callback from Google, which means the user is already authorized, we just validate the existing user to create a new record, then we could return tokens for accessing other resources 💭

spec/models/user_spec.rb
olivierobert: Why does it need to be a YAML file? 🤔  We can load any file type, including JSON:

```
File.read(File.join('spec', 'fixtures', 'public', filename)) 
```

hoangnguyen92dn: @olivierobert We can load the content file, but when we put the content into the `OmniAuth::AuthHash.new` to create a new auth_hash, it will throw an error. Using `eval(file_content)` can resolve the issue unfortunately Rubocop shows a warning about security risks.  

```
NoMethodError:
       undefined method `each_pair' for "{  :provider => \"google_oauth2\",  :uid => \"123456789\",  :info => {    :name => \"John Doe\",    :email => \"john.doe@example.com\",    :first_name => \"John\",    :last_name => \"Doe\",    :image => \"https://lh3.googleusercontent.com/url/photo.jpg\"  },  :credentials => {      :token => \"token\",      :refresh_token => \"another_token\",      :expires_at => 1354920555,      :expires => true  },  :extra => {    :raw_info => {      :sub => \"123456789\",      :email => \"john.doe@example.com\",      :email_verified => true,      :name => \"John Doe\",      :given_name => \"John\",      :family_name => \"Doe\",      :profile => \"https://plus.google.com/123456789\",      :picture => \"https://lh3.googleusercontent.com/url/photo.jpg\",      :gender => \"male\",      :birthday => \"0000-06-25\",      :locale => \"en\",      :hd => \"example.com\"    },    :id_info => {      \"iss\" => \"accounts.google.com\",      \"at_hash\" => \"HK6E_P6Dh8Y93mRNtsDB1Q\",      \"email_verified\" => \"true\",      \"sub\" => \"10769150350006150715113082367\",      \"azp\" => \"APP_ID\",      \"email\" => \"jsmith@example.com\",      \"aud\" => \"APP_ID\",      \"iat\" => 1353601026,      \"exp\" => 1353604926,      \"openid_id\" => \"https://www.google.com/accounts/o8/id?id=ABCdfdswawerSDFDsfdsfdfjdsf\"    }  }}":String
       Did you mean?  each_char
     # ./spec/models/user_spec.rb:12:in `new'
     # ./spec/models/user_spec.rb:12:in `block (4 levels) in <top (required)>'
     # ./spec/support/vcr.rb:31:in `block (2 levels) in <top (required)>'
```

Anw, we will update to use Faker to generate the hash instead of reading from the file 😄

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Users
+      class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+        respond_to :json
+        def google_oauth2
```
hoangmirs, line 8: Please add a line break here
```suggestion
        respond_to :json
 
        def google_oauth2
```

config/routes.rb
```diff
@@ -7,6 +7,10 @@
     namespace :v1 do
       resources :upload, only: [:create], controller: 'search_results'
 
+      devise_for :users, controllers: {
+        omniauth_callbacks: 'api/v1/users/omniauth_callbacks'
+       }
```
hoangmirs, line 12:
```suggestion
      }
```

hoangmirs (approved):

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        end
+
+        private
+
+        def auth
+          @auth ||= request.env['omniauth.auth']
+        end
+
+        def render_success
+          render json: { success: true }, status: :ok
```
olivierobert, line 26: Never just return a strong or boolean Here the user resource can be returned.

app/models/user.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def set_user_attributes(user, auth)
+      user.provider = auth.provider
+      user.uid = auth.uid
+
+      profile = auth.info
+      user.email = profile.email
+      user.password = Devise.friendly_token[0, 20]
+      user.full_name = profile.name
+      user.avatar_url = profile.image
```
olivierobert, line 25: What is this is null? Not all user accounts have a profile image. You might need to use the [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) if that can cause an error 💡

spec/requests/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Omniauth Callbacks', type: :request do
+  describe 'POST #google_oauth2' do
+    context 'given the email is available' do
+      it 'renders a successful response' do
+        user = Fabricate(:user)
+        Request::OmniAuthCallback.new(:google_oauth2).mock uid: user.uid
```
olivierobert, line 10: The usage of parenthesis is inconsistent 🙈 Why omit them for the second method call? 

```suggestion
        Request::OmniAuthCallback.new(:google_oauth2).mock(uid: user.uid)
```

olivierobert (changes requested):

config/routes.rb
hoangnguyen92dn: 🙈 The 'Format Document' doesn't work at all.

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
hoangnguyen92dn: @olivierobert My intention is about returning tokens on success, so this is a temporary response 😅

olivierobert (approved):

Tuubz (approved):

Issue #18, opened 2023-05-25
[Backend] As a user, I can see the details of a search result


Tuubz:
## Why

The user can see the details of a search result

## Acceptance Criteria

- A user can only see the details of the search results he uploaded, based on the authorization token

- Create a new endpoint: `GET api/v1/search_results/{searchResultId}`

    - **Header**: Authorization (bearer token)
    -  **Request**: `searchResultId`
    - **Response**: List the details of a search result
    
- Write unit and integration tests
- Handle errors: Unauthorized, ...

## Design
N/A

## Resources
N/A

PR #44, opened 2023-06-27
[#6] [Backend] As a user, I can sign-up with e-mail and password

### LINKED ISSUE #6: [Backend] As a user, I can sign-up with e-mail and password
## Why

Allow a user to register a new account, once authenticated they can interact with the application.

## Acceptance Criteria

- Create a table for users, which must have the following fields:

   - E-mail → unique, also add validation (ex: must contain @)
   - Password → encrypted, also add validation (ex: must contain at least 8 characters)

**Note**: Make use of `devise` to handle the users

- Create a new endpoint: `POST api/v1/sign_up`:

   - **Request**: E-mail, password, client ID & client secret
   - **Response**: Success message + [HTTP code](ttps://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses)

- Handle errors

## Design
N/A

## Resources
- https://github.com/heartcombo/devise
- https://github.com/doorkeeper-gem/doorkeeper

Image

toby-thanathip:
Close #6

## What happened 👀

- Added sign-up for email/password (the provider can be either `email` or `google_oauth2`)

For the sake of simplicity, I've decided to follow [@olivierobert's suggestion](https://nimble-co.slack.com/archives/C058GEPCT7F/p1688700310152069?thread_ts=1688634166.293839&cid=C058GEPCT7F).

*This means there can only be one user for one e-mail*, an error message will be returned otherwise.

- Also added `client_id` & `client_secret` authorization through `doorkeeper` (they are required in the body request)

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/94153688-006d-4718-a48f-a20b661d8215

### DIFF
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -26,10 +26,10 @@ module Api
       end
 
       def render_success(search_results)
-        data = SearchResultsSerializer.new(search_results).serializable_hash[:data]
-        message = I18n.t('activemodel.notices.models.search_result.create')
+        success_message = I18n.t('activemodel.notices.models.search_result.create')
+        search_results_data = SearchResultsSerializer.new(search_results, meta: { message: success_message })
 
-        render json: { data: data, meta: { message: message } }, status: :created
+        render json: search_results_data, status: :created
       end
 
       def render_failed(upload_form)
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -33,6 +33,7 @@ module Api
       end
 
       def render_failed(upload_form)
+        # TODO: return error messages in JSON:API format
         render json: { errors: upload_form.errors.full_messages }, status: :unprocessable_entity
       end
     end
--- a/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
@@ -4,14 +4,11 @@ module Api
   module V1
     module Users
       class OmniauthCallbacksController < Devise::OmniauthCallbacksController
-        rescue_from Exception, with: :render_error
-
         respond_to :json
 
         def google_oauth2
           user = User.from_omniauth(auth)
-
-          render_success(user)
+          user.errors.empty? ? render_success(user) : render_error(user)
         end
 
         private
@@ -27,8 +24,8 @@ module Api
           render json: { data: token_data }, status: :ok
         end
 
-        def render_error(exception)
-          render json: { error: exception.message }, status: :internal_server_error
+        def render_error(user)
+          render json: { errors: user.errors.full_messages }, status: :internal_server_error
         end
       end
     end
--- a/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
@@ -8,7 +8,11 @@ module Api
 
         def google_oauth2
           user = User.from_omniauth(auth)
-          user.errors.empty? ? render_success(user) : render_error(user)
+          if user.errors.empty?
+            render_success(user)
+          else
+            render_error(user)
+          end
         end
 
         private
@@ -19,9 +23,9 @@ module Api
 
         def render_success(user)
           oauth_token = OauthToken.generate_access_token(user)
-          token_data = OauthTokenSerializer.new(oauth_token).serializable_hash[:data]
+          token_data = OauthTokenSerializer.new(oauth_token)
 
-          render json: { data: token_data }, status: :ok
+          render json: token_data, status: :ok
         end
 
         def render_error(user)
--- a/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
@@ -8,6 +8,7 @@ module Api
 
         def google_oauth2
           user = User.from_omniauth(auth)
+
           if user.errors.empty?
             render_success(user)
           else
--- a/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/api/v1/users/omniauth_callbacks_controller.rb
@@ -29,6 +29,7 @@ module Api
         end
 
         def render_error(user)
+          # TODO: return error messages in JSON:API format
           render json: { errors: user.errors.full_messages }, status: :internal_server_error
         end
       end
--- a/app/controllers/api/v1/users/registrations_controller.rb
+++ b/app/controllers/api/v1/users/registrations_controller.rb
@@ -31,10 +31,14 @@ module Api
         end
 
         def render_error(user)
+          # TODO: return error messages in JSON:API format
+
           render json: { errors: user.errors.full_messages }, status: :unprocessable_entity
         end
 
         def ensure_valid_client
+          # TODO: return error messages in JSON:API format
+
           @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
 
           render json: { errors: I18n.t('doorkeeper.errors.messages.invalid_client') }, status: :forbidden if @client_app.blank?
--- a/app/controllers/api/v1/users/registrations_controller.rb
+++ b/app/controllers/api/v1/users/registrations_controller.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Users
+      class RegistrationsController < Devise::RegistrationsController
+        respond_to :json
+
+        before_action :ensure_valid_client
+
+        def create
+          user = User.from_email(user_params[:email], user_params[:password])
+          user.errors.empty? ? render_success(user) : render_error(user)
+        end
+
+        private
+
+        def user_params
+          params.require(:user).permit(:email, :password)
+        end
+
+        def render_success(user)
+          data = UserSerializer.new(user).serializable_hash[:data]
+          message = I18n.t('activemodel.notices.models.user.create')
+
+          render json: { data: data, meta: { message: message } }, status: :created
+        end
+
+        def render_error(user)
+          render json: { errors: user.errors.full_messages }, status: :unprocessable_entity
+        end
+
+        def ensure_valid_client
+          @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
+
+          render json: { errors: I18n.t('doorkeeper.errors.messages.invalid_client') }, status: :forbidden if @client_app.blank?
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/users/registrations_controller.rb
+++ b/app/controllers/api/v1/users/registrations_controller.rb
@@ -10,6 +10,7 @@ module Api
 
         def create
           user = User.from_email(user_params[:email], user_params[:password])
+
           if user.errors.empty?
             render_success(user)
           else
@@ -39,9 +40,9 @@ module Api
         def ensure_valid_client
           # TODO: return error messages in JSON:API format
 
-          @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
+          client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
 
-          render json: { errors: I18n.t('doorkeeper.errors.messages.invalid_client') }, status: :forbidden if @client_app.blank?
+          render json: { errors: I18n.t('doorkeeper.errors.messages.invalid_client') }, status: :forbidden if client_app.blank?
         end
       end
     end
--- a/app/controllers/api/v1/users/registrations_controller.rb
+++ b/app/controllers/api/v1/users/registrations_controller.rb
@@ -10,7 +10,11 @@ module Api
 
         def create
           user = User.from_email(user_params[:email], user_params[:password])
-          user.errors.empty? ? render_success(user) : render_error(user)
+          if user.errors.empty?
+            render_success(user)
+          else
+            render_error(user)
+          end
         end
 
         private
@@ -20,10 +24,10 @@ module Api
         end
 
         def render_success(user)
-          data = UserSerializer.new(user).serializable_hash[:data]
-          message = I18n.t('activemodel.notices.models.user.create')
+          success_message = I18n.t('activemodel.notices.models.user.create')
+          user_data = UserSerializer.new(user, meta: { message: success_message })
 
-          render json: { data: data, meta: { message: message } }, status: :created
+          render json: user_data, status: :created
         end
 
         def render_error(user)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -17,24 +17,47 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :validatable,
          :doorkeeper, :omniauthable, omniauth_providers: [:google_oauth2]
 
+  enum provider: { email: 'email', google_oauth2: 'google_oauth2' }
+
+  validates :provider, inclusion: { in: providers.keys }
+
   class << self
-    def from_omniauth(auth)
-      User.find_or_create_by(provider: auth.provider, uid: auth.uid, email: auth.info.email) do |user|
-        set_user_attributes(user, auth)
-      end
+    def from_email(email, password)
+      user = find_or_initialize_user(
+        email: email,
+        password: password,
+        provider: User.providers[:email]
+      )
+
+      user.persisted? ? user.errors.add(:email, :taken) : user.save
+
+      user
     end
 
-    private
+    def from_omniauth(auth)
+      user = find_or_initialize_user(
+        email: auth.info.email,
+        password: Devise.friendly_token[0, 20],
+        provider: auth.provider,
+        auth: auth
+      )
 
-    def set_user_attributes(user, auth)
-      user.provider = auth.provider
-      user.uid = auth.uid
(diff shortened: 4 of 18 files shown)


olivierobert (commented):
While this implementation works, it creates a difference between logging in with Google and Email + Password (i.e., the latter does not go through omniauth). It would be best to create [an Email + Password strategy with Omniauth](https://github.com/omniauth/omniauth/wiki/Strategy-Contribution-Guide). Check out this SO response for more context: https://stackoverflow.com/a/7255898

Thankfully there is a gem that you can leverage: https://github.com/omniauth/omniauth-identity

db/migrate/20230705085150_remove_provider_from_users.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveProviderFromUsers < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :users, :provider, :string
+    remove_column :users, :uid, :string
+    remove_column :users, :full_name, :string
+    remove_column :users, :avatar_url, :string
```
olivierobert, line 6: `avatar_url` should remain as it can be updatable in the user profile

db/migrate/20230629073203_create_identities.rb
```diff
@@ -0,0 +1,12 @@
+class CreateIdentities < ActiveRecord::Migration[7.0]
+  def change
+    create_table :identities do |t|
+      t.references :user, null: false, foreign_key: true
+      t.string :email, null: false
+      t.string :password_digest, null: false
+      t.string :provider, null: false
```
olivierobert, line 7: We should have the following:

```
provider_uid -> nullable but to store third-party IDs
identifier -> non-nullable  as it can be email or any other value used by the identify provider 
credential -> nullable as it will be empty for email+password auth (for Facebook we can use that token to fetch user data) 
```

Tuubz: @olivierobert I tried removing `password_digest`, but it results to an error unless I add it 💭

hoangmirs: I think you need to modify this code as well
https://github.com/nimblehq/ic-rails-huey-toby/pull/44/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR30

Tuubz: @hoangmirs Yes, correct. I removed it there as well 🙏 But still the same error

Screen Shot 2023-07-06 at 17 20 22

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
       private
 
       def upload_form_params
-        params.permit(:search_engine, :csv_file)
+        params.permit(:search_engine, :csv_file, :client_id, :client_secret)
```
olivierobert, line 25: `:client_id` and `:client_secret` are only sent for the sign up endpoint. After that the access token is used to authorize every request 🔑

olivierobert (commented):
I do NOT recommend combining sign-in and sign-up. Prefer to keep these separate. Sign-up should be done first.

app/controllers/api/v1/search_results_controller.rb
Tuubz: @olivierobert I will also add it to the sign-in endpoint, as the response for sign-up will not include the token yet (because we must verify the user's email first) 🙏

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        private
+
+        def user_params
+          params.require(:user).permit(:email, :password)
+        end
+
+        def render_success(user)
+          data = UserSerializer.new(user).serializable_hash[:data]
+          message = I18n.t('activemodel.notices.models.user.create')
```
hoangnguyen92dn, line 24: Please explicit the name to clarify the purpose. 😅 

```suggestion
          user_data = UserSerializer.new(user).serializable_hash[:data]
          meta_message = I18n.t('activemodel.notices.models.user.create')
```

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+          render json: { data: data, meta: { message: message } }, status: :created
+        end
+
+        def render_error(user)
+          render json: { errors: user.errors.full_messages }, status: :unprocessable_entity
```
hoangmirs, line 30: The current error responses are incorrect 🙈 . The format should follow the [JSON:API specs](https://jsonapi.org/format/#document-top-level):

> errors: an array of [error objects](https://jsonapi.org/format/#errors).

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        def create
+          user = User.from_email(user_params[:email], user_params[:password])
+          user.errors.empty? ? render_success(user) : render_error(user)
+        end
+
+        private
+
+        def user_params
+          params.require(:user).permit(:email, :password)
```
hoangnguyen92dn, line 19: Would it be possible to flatten the params? 🤔 I mean get rid of unnecessary `:user`.

app/models/user.rb
```diff
@@ -17,24 +17,47 @@ class User < ApplicationRecord
...
-        set_user_attributes(user, auth)
-      end
+    def from_email(email, password)
+      user = find_or_initialize_user(
+        email: email,
+        password: password,
+        provider: User.providers[:email]
+      )
+
+      user.persisted? ? user.errors.add(:email, :taken) : user.save
```
hoangmirs, line 32: If the `email` field is unique, I think we don't need this line, as it will be handled by the User validation 🤔

app/models/user.rb
```diff
@@ -17,24 +17,47 @@ class User < ApplicationRecord
...
+        email: auth.info.email,
+        password: Devise.friendly_token[0, 20],
+        provider: auth.provider,
+        auth: auth
+      )
 
-    def set_user_attributes(user, auth)
-      user.provider = auth.provider
-      user.uid = auth.uid
+      user.provider == User.providers[:email] ? user.errors.add(:email, :taken) : user.persisted? || user.save
```
hoangmirs, line 45: Do we need to check the `email` providers in this case? You are no longer using the `omniauth-identity` gem, right?

app/serializers/user_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class UserSerializer
+  include JSONAPI::Serializer
```
hoangmirs, line 4: We can create an `ApplicationSerializer` file and put this into it. Then you can let this serializer inherits it

spec/requests/registrations_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Registrations', type: :request do
+  describe 'POST #create' do
+    context 'given the email is NOT already taken' do
```
hoangmirs, line 7: As this is the happy case, should we rename this test to:
```suggestion
    context 'given the email is available' do
```

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -27,8 +24,8 @@ def render_success(user)
           render json: { data: token_data }, status: :ok
         end
 
-        def render_error(exception)
-          render json: { error: exception.message }, status: :internal_server_error
+        def render_error(user)
```
hoangnguyen92dn, line 27: Self-note: Update to use method `render_error` added on the `error_handler` concern

https://github.com/nimblehq/ic-rails-huey-toby/pull/46/commits/d9f38283b49d6bf8a88c377b659ada64dbadfb69#diff-9b9c5d5d6a6266cc2ae1139314a3a72eaa8367b34eee0a66d45b0bcfc86e6172

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        def render_error(user)
+          render json: { errors: user.errors.full_messages }, status: :unprocessable_entity
+        end
+
+        def ensure_valid_client
+          @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
+
+          render json: { errors: I18n.t('doorkeeper.errors.messages.invalid_client') }, status: :forbidden if @client_app.blank?
+        end
```
hoangnguyen92dn, line 37: Can we have the validation in concern instead, so we could reuse it in other controllers? 😄 

Similar to this: https://github.com/nimblehq/nimble-survey-web/blob/develop/app/controllers/concerns/api/v1/oauth_application_verifiable.rb#L5

app/serializers/user_serializer.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  attributes :email, :full_name, :avatar_url, :provider, :provider_uid
```
hoangnguyen92dn, line 6: For users, `provider` and `provider_uid` seems redundant IMO.

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  module V1
+    module Users
+      class RegistrationsController < Devise::RegistrationsController
+        respond_to :json
+
+        before_action :ensure_valid_client
+
+        def create
+          user = User.from_email(user_params[:email], user_params[:password])
+          user.errors.empty? ? render_success(user) : render_error(user)
```
olivierobert: The ternary operator is rarely used in Ruby, i.e., it is not idiomatic (and also i would be inconsistent with Huey's implementation):

```suggestion
          if user.errors.empty?
            render_success(user) :
          else
             render_error(user)
          end
```

or using early returns (but less preferred as it can cause weird issues in controllers 😆 ):

```suggestion
          return render_error(user) unless user.errors.empty?
          
          ender_success(user)
```

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+
+        private
+
+        def user_params
+          params.require(:user).permit(:email, :password)
+        end
+
+        def render_success(user)
+          data = UserSerializer.new(user).serializable_hash[:data]
```
olivierobert, line 23: Avoid using `data` as a variable name 🙈 

```suggestion
          user_data = UserSerializer.new(user).serializable_hash[:data]
```

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+          message = I18n.t('activemodel.notices.models.user.create')
+
+          render json: { data: data, meta: { message: message } }, status: :created
```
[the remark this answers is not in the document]
Tuubz: Thank you, I'll keep the message so it's aligned with `search_results_controller.rb`

In terms of `serializable_hash`, I think it's necessary if we only want the fields inside `data`

Otherwise it will look like this (we don't want `data` inside `data`):

Screen Shot 2023-07-11 at 14 09 36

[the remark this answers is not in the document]
hoangmirs: Then you can remove the root data attribute 😄 
```suggestion
          message = I18n.t('activemodel.notices.models.user.create')
          user_serializer = UserSerializer.new(user, meta: { message: message })

          render json:user_serializer, status: :created
```

app/controllers/api/v1/users/registrations_controller.rb
Tuubz: Hmmm, which one of these fields do you recommend us to return? 💭 cc @hoangnguyen92dn 

Screen Shot 2023-07-11 at 14 21 06

app/models/user.rb
Tuubz: Since we're using `find_or_initialize_user`, we need `user.persisted?` to know if the record already existed or not.

Without this line, it will just simply return the existing user, without giving any error

app/models/user.rb
Tuubz: Yes, we do. 🙏 

Let's say the user clicks the google sign-in button:

- If a user already had a `google_oauth2` account, then simply return the same user (no error)
- If a user already had an `email` account, then return an error that the e-mail already has been taken

app/controllers/api/v1/users/registrations_controller.rb
Tuubz: @hoangmirs Oh, I think @hoangnguyen92dn has provided me the answer [here](https://github.com/nimblehq/ic-rails-huey-toby/pull/44#discussion_r1259156268) 👍

app/serializers/user_serializer.rb
Tuubz: Hmmmm, are you suggesting we should remove it from this serializer class or from the table? 💭

app/controllers/api/v1/users/registrations_controller.rb
Tuubz: @hoangnguyen92dn Should I add a TODO comment and leave it for you to update it? 🙏

app/serializers/user_serializer.rb
Tuubz: At this stage, I guess we don't need `provider` at all

However, if we think ahead and start adding `twitter` for example then we should keep it 👍

app/controllers/api/v1/users/registrations_controller.rb
hoangnguyen92dn: @Tuubz That's fine. We could update in the next upcoming tasks 👍

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
Tuubz: Related to https://github.com/nimblehq/ic-rails-huey-toby/pull/44#discussion_r1259362567

app/serializers/user_serializer.rb
hoangnguyen92dn: @Tuubz I meant for the response we don't need those fields 😅 For now, I think we could leave it as it is 👌

app/serializers/application_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ApplicationSerializer
+  include JSONAPI::Serializer
```
hoangnguyen92dn, line 4: @Tuubz Do you think we should have concerns for the serializer and moving render method for success and error inside? 🤔 If so I will refactor on PR #18

hoangnguyen92dn (approved):
LGTM 🚀

app/serializers/application_serializer.rb
Tuubz: Sounds good to me 👍

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -4,14 +4,15 @@ module Api
   module V1
     module Users
       class OmniauthCallbacksController < Devise::OmniauthCallbacksController
-        rescue_from Exception, with: :render_error
-
         respond_to :json
 
         def google_oauth2
           user = User.from_omniauth(auth)
```
olivierobert, line 10: A lin ebreak would nice to separate the variable definition and conditional block

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+module Api
+  module V1
+    module Users
+      class RegistrationsController < Devise::RegistrationsController
+        respond_to :json
+
+        before_action :ensure_valid_client
+
+        def create
+          user = User.from_email(user_params[:email], user_params[:password])
```
olivierobert, line 12: Ditto here about having a line break

app/controllers/api/v1/users/registrations_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        def render_error(user)
+          # TODO: return error messages in JSON:API format
+
+          render json: { errors: user.errors.full_messages }, status: :unprocessable_entity
+        end
+
+        def ensure_valid_client
+          # TODO: return error messages in JSON:API format
+
+          @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
```
olivierobert, line 42: Why does this need to be an instance variable? WOuldn't the following work?

```suggestion
          client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
```

app/models/user.rb
olivierobert: The concern raised by Hoang comes from the fact that team rarely uses `find_or_initialize_user` 😅 Also, you have not revisited this implementation based on an earlier comment about the ternary operator not being idiomatic in Ruby

app/models/user.rb
olivierobert: The code is too complex now 🙈 Ternary operator and another `||` operand 🤯 You need to find a simpler way to handle such complex logic.

app/serializers/application_serializer.rb
olivierobert: Guys, stop moving everything into concerns. It makes the code harder to read and discover ⚠️ Only use concerns for complex logic that needs to be shared. Defining methods on the application controller would work better for rendering logic (i.e, there is nothing complex)

olivierobert (changes requested):

app/models/user.rb
Tuubz: Thank you guys, I was indeed overcomplicating it a lot 🙈 

I have simplified the code, looks a lot better now  🚀

olivierobert (approved):

hoangmirs (approved):

PR #45, opened 2023-07-03
[#11][#22] [Backend] As a user, I must be authorized to upload keywords

### LINKED ISSUE #11: [Backend] As a user, I can sign-in with Google

### LINKED ISSUE #22: [Backend] As a user, I must be authorized to upload keywords
## Why

A user must be authorized to upload a CSV file with keywords 

## Acceptance Criteria

- Add references with the `users` table (one user has many search results & one search result has one user)
- Add a header field for the existing endpoint: `POST api/v1/upload?engine={google/bing}`

**Note**: Check with doorkeeper if the access token is valid

## Design
N/A

## Resources
N/A

hoangnguyen92dn:
- Close #11 
- Close #22 

## What happened 👀

- Add requires authorization for the existing endpoint to upload the search file

## Insight 📝

- Add (doorkeeper)[https://github.com/doorkeeper-gem/doorkeeper] to accept authorized requests with the token on the header.
  - Since `doorkeeper` creates tables to manage the token with the user's id, there is a migration to update the user's id to uuid to match with `oauth_access_tokens` table.
- Update the Omniauth callback to return access_token for sign-in and sign-up cases.
- Update UnitTest
  - Because the `search_result_controller.rb` now requires the token to access, the tests request needs to be updated to include the authorization header.

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/35295d6b-b5e7-4dd8-a7ef-09043fc74a54


app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -23,12 +18,17 @@ def auth
           @auth ||= request.env['omniauth.auth']
         end
 
-        def render_success
-          render json: { success: true }, status: :ok
-        end
+        def render_response(user)
+          oauth_token = OauthToken.generate_access_token(user)
+          data = OauthTokenSerializer.new(oauth_token).serializable_hash[:data]
```
hoangmirs, line 23: Should we explicit this: `token_data`?
```suggestion
          token_data = OauthTokenSerializer.new(oauth_token).serializable_hash[:data]
```

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -23,12 +18,17 @@ def auth
...
+
+          message = if user.persisted?
```
hoangmirs, line 25: Do we need this message? As this is an API, so I think this message is redundant. We often use it to show in the view - [flash message](https://www.rubyguides.com/2019/11/rails-flash-messages/#:~:text=A%20flash%20message%20is%20a,User%20not%20found%E2%80%9D%20(error))

db/migrate/20230704091131_update_id_users.rb
```diff
@@ -0,0 +1,16 @@
+class UpdateIdUsers < ActiveRecord::Migration[7.0]
```
hoangmirs, line 1: Should we change the name to:
```suggestion
class UpdateIdColumnOfUsers < ActiveRecord::Migration[7.0]
```
This is recommended by [Database section on Compass](https://www.rubyguides.com/2019/11/rails-flash-messages/#:~:text=A%20flash%20message%20is%20a,User%20not%20found%E2%80%9D%20(error))

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
olivierobert: `data` is never a good variable 🙃

app/models/oauth_token.rb
```diff
@@ -0,0 +1,26 @@
...
+        refresh_token: Doorkeeper::OAuth::Helpers::UniqueToken.generate,
+        expires_in: Doorkeeper.configuration.access_token_expires_in.to_i,
+        scopes: Doorkeeper.configuration.default_scopes
+      )
+    end
+
+    private
+
+    def generate_application
+      Doorkeeper::Application.find_or_create_by(name: OAUTH_APPLICATION_NAME,
```
olivierobert, line 22: The OAuth application should be generated via the seed file when setting up the application. It should rely on having a user trying to authenticate for the first time 🙈 Also. ensuring that the OAuth application exists upon deployment is needed. So `find_or_create_by` is not the right call here. Also, the method name does not describe what it does.

app/models/oauth_token.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class OauthToken < ApplicationRecord
+  OAUTH_APPLICATION_NAME = 'ic-rails-huey-toby'
+  OAUTH_APPLICATION_REDIRECT_URI = 'urn:ietf:wg:oauth:2.0:oob'
+
+  class << self
+    # https://github.com/doorkeeper-gem/doorkeeper/issues/846#issuecomment-230297646
```
olivierobert, line 8: Let's add a comment to avoid all devs to have to read this issue 🙈 

```suggestion
		# Generate token manually to automatically log a user in after registering    
		# https://github.com/doorkeeper-gem/doorkeeper/issues/846#issuecomment-230297646
```

olivierobert (changes requested):

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
hoangnguyen92dn: @hoangmirs I added the message to determine the case user sign in or sign up 😄 Would it be possible to keep it here? 😅

app/models/oauth_token.rb
hoangnguyen92dn: @olivierobert Thanks for your comment. I added the default oauth application to the seed file, besides that, the new migration file was also added to make sure a new record will be added to oauth_application table on deployment. So in the `generate_access_token` method, the first record of `Doorkeeper::Application` will be always used for generating access token. I am missing anything? 😄

db/seeds.rb
```diff
@@ -5,3 +5,22 @@
...
+ secret = "1RuVgBFgU24MTp2ZUYriT0m7I0CaLNE3BvPRRhDMfSYBsYt6x3la6POHEp6WwtQ2"
+when "development"
+  redirect_uri  = "https://localhost:3000/api/v1/users/auth/google_oauth2/callback"
+  uid = "MouWUcR81F7QDdgzwVlvSqonftQ8H1iszHJqoiuSUnJVvN0eCjTO9dbZJmKPUXlb"
+  secret = "MA4gWSjuDF1mcZ60l2NqC7YC86h1jRDEvjnq13QN09hwWXSsJzT6ZYv5mZMNnQcO"
+end
+
+if Doorkeeper::Application.count.zero?
+    Doorkeeper::Application.create(name: 'ic-rails-huey-toby-client', redirect_uri: redirect_uri, uid: uid, secret: secret)
+end
```
olivierobert, line 26: Missing EOF line break

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
olivierobert: It is a security concern. Hackers could use the API message to check if a user account exists. It is best to remove any such information.

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -4,17 +4,13 @@ module Api
   module V1
     module Users
       class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+        rescue_from Exception, with: :render_error
```
olivierobert, line 7: WHY U NO LIKE adding line breaks 😆 

```suggestion
        rescue_from Exception, with: :render_error
        
```

olivierobert (approved):
Just some minor formatting issues. Pre-approving ✅

app/models/oauth_token.rb
olivierobert: All good now 👍

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
hoangnguyen92dn: 🙈

hoangmirs (approved):

hoangmirs (approved):

Tuubz (approved):

Tuubz:
@olivierobert Could I have your last approval? 🙇

PR #46, opened 2023-07-06
[#13][Backend] As a user, I can see the list of search results

### LINKED ISSUE #13: [Backend] As a user, I can see the list of search results
## Why

The user can see all the search results

## Acceptance Criteria

- A user can only see the search results he uploaded, based on the authorization token

- Create a new endpoint: `GET api/v1/search_results?page[number]=1&page[size]=2`

    -  **Request**: `page[number]` & `page[size]`
    
**Note**: Use "pagy" for pagination

## Design
N/A

## Resources
https://github.com/ddnexus/pagy

hoangnguyen92dn:
- Close #13 

## What happened 👀

- Create a new endpoint: `GET api/v1/search_results?page[number]=1&page[size]=2`
    -  **Request**: `page[number]` & `page[size]`

## Insight 📝

- Use `pagy` to handle the pagination
- Try to setup authorization header for all test request similar to [reference on The1
](https://github.com/nimblehq/the1-api/blob/37a682553787e5c0ee96da57e2fa5405a8da01c5/spec/support/authentication_helpers.rb#L6-L12) unfortunately it doesn't work at all cause of the `request` always `nil`, I decided to move back by creating new method to create authorization header and call on every request.
- Add UnitTest

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/a67cd1ba-0bde-47f6-bc30-b79ab2e0346d


app/controllers/api/v1/search_results_controller.rb
```diff
@@ -21,6 +18,16 @@ def create
 
       private
 
+      def create_upload_form
+        current_user = User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
```
olivierobert, line 22: The usual way to do this is to define a method `current_user` in ApplicationController so that it can be re-used in all controllers. Example: https://github.com/nimblehq/nimble-survey-web/blob/3f8c20fdf704f7fb502f236a5087bd8500ca9fcd/app/controllers/concerns/api/v1/doorkeeper_authentication.rb#L16

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -3,13 +3,8 @@
...
   module V1
     class SearchResultsController < ApplicationController
-      before_action :doorkeeper_authorize!
-
       def create
-        upload_form = UploadForm.new(
-          search_engine: upload_form_params[:search_engine],
-          csv_file: upload_form_params[:csv_file]
-        )
+        upload_form = create_upload_form
```
olivierobert, line 7: This is an anti-pattern which I have raised before. Prefer the following:

```
def creates
  if upload_form.valid?
		##...
end 

private

def upload_form
 @upload_form ||= UploadForm.new(
 #...
 )
end
```

This operator will return the existing instance if it is already defined. 

In Ruby, the principle "[tell, don't ask](https://thoughtbot.com/blog/tell-dont-ask)" should be followed.

olivierobert (changes requested):

app/controllers/concerns/pagination.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  included do
+    def paginated_resources_for(resources, **options)
+      pagy resources, pagination_params.merge(options)
+    end
+
+    def pagination_params
+      {
+        page: params.dig(:page, :number) || Pagy::DEFAULT[:page],
+        items: params.dig(:page, :size) || Pagy::DEFAULT[:items]
```
hoangmirs, line 16: Do you know if we need to specify the default values? Should it be used when we pass params as nil?
```suggestion
        page: params.dig(:page, :number),
        items: params.dig(:page, :size)
```

db/migrate/20230706033605_add_user_id_to_search_results.rb
```diff
@@ -0,0 +1,12 @@
+class AddUserIdToSearchResults < ActiveRecord::Migration[7.0]
+  def up
+    # Add user_id columns
+    add_column :search_results,    :user_id, :uuid, null: true
+
+    add_foreign_key :search_results, :users, column: :user_id
```
hoangmirs, line 6: Can we use `add_reference` to replace these lines?
```suggestion
    add_reference   :search_results, :users, foreign_key: true
```

app/controllers/concerns/pagination.rb
hoangnguyen92dn: @hoangmirs Thanks for spotting out. It requires when I extract to form objects because it throws the exception pagy undefined.

db/migrate/20230706033605_add_user_id_to_search_results.rb
hoangnguyen92dn: @hoangmirs 👍 I will need to add `type: :uuid` since the [default type is int ](https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference)so it doesn't match with user_id type.

```
add_reference   :search_results, :users, type: :uuid, foreign_key: true
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -19,8 +14,28 @@ def create
...
 
+      def index
+        search_result_list = SearchResult.where(user_id: current_user.id).order(:id)
+        pagy, paginated_results = paginated_resources_for(search_result_list)
+
+        search_result_data = SearchResultsSerializer.new(paginated_results).serializable_hash[:data]
+        pagy_meta = meta_from_pagy(pagy)
+
+        render json: { data: search_result_data, meta: pagy_meta }, status: :ok
+      rescue Pagy::OverflowError
```
olivierobert, line 25: That's odd and unexpected. IS this really required? We do not need to catch every error. Any unhandled error will return a server error which is better than rendering 404 when it is not about not finding a record.

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -19,8 +14,28 @@ def create
         end
       end
 
+      def index
+        search_result_list = SearchResult.where(user_id: current_user.id).order(:id)
```
olivierobert, line 18: Not needed now but the team usually uses https://github.com/varvet/pundit to automatically filter and authorize current user actions 💡

db/migrate/20230706033605_add_user_id_to_search_results.rb
hoangnguyen92dn: @hoangmirs I noticed the default column will be named `users_id` instead of `user_id`, it will affect the reference in the codebase and test classes, so I think it'd be better if we revert the update 🙏

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -35,6 +50,15 @@ def render_success(search_results)
       def render_failed(upload_form)
         render json: { errors: upload_form.errors.full_messages }, status: :unprocessable_entity
       end
+
+      def meta_from_pagy(pagy)
```
olivierobert, line 54: Couldn't this be moved to the pagination concern?

db/migrate/20230706033605_add_user_id_to_search_results.rb
hoangnguyen92dn: @hoangmirs 🙈 The reference_name should be `:user` instead of `:users`. Pushed 🚀

app/controllers/concerns/error_handler.rb
Tuubz, line 1: Not sure, but should this class contain the keyword `concern`? 

- `ErrorHandler` to `ErrorHandlerConcern`
- `DoorkeeperAuthentication` to `DoorkeeperAuthenticationConcern`

hoangnguyen92dn: @Tuubz I am not sure about the convention our team is following for concerns. Seems it should be named with `able` suffix. cc @hoangmirs @olivierobert

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -19,8 +14,26 @@ def create
         end
       end
 
+      def index
+        # TODO: Add pundit to filter search results by user
+        search_result_list = SearchResult.where(user_id: current_user.id).order(:id)
```
Tuubz, line 19: Here's a cooler way of doing it 🤩 
```suggestion
        search_result_list = current_user.search_results.order(:id)
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -19,8 +14,26 @@ def create
...
+        pagy, paginated_results = paginated_resources_for(search_result_list)
+
+        search_result_serializer = SearchResultsSerializer.new(paginated_results, meta: meta_from_pagy(pagy))
```
Tuubz, line 22: Do you it would be a good idea to refactor it with `render_success`? 💭

app/controllers/concerns/doorkeeper_authentication.rb
```diff
@@ -0,0 +1,27 @@
...
+  include ErrorHandler
+
+  included do
+    before_action :doorkeeper_authorize!
+  end
+
+  private
+
+  def current_user
+    User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
```
Tuubz, line 14: Interesting, I thought `current_user` was a built-in function from `devise` 💭

Tuubz (commented):
Cool 🚀

app/controllers/concerns/doorkeeper_authentication.rb
Tuubz: Chat GPT 🤖 

> The current_user method is typically provided by Devise in the context of web applications, where session management is handled through cookies. However, in an API context, the session management is often handled differently, such as using tokens or API keys.

> If you're building an API and want to authenticate users with Devise, you'll typically use a different approach to track the authenticated user. For example, you might use token-based authentication, such as JSON Web Tokens (JWT) or OAuth.

app/controllers/api/v1/search_results_controller.rb
hoangnguyen92dn: @Tuubz Looks better 👍

app/models/search_result.rb
```diff
@@ -4,6 +4,8 @@ class SearchResult < ApplicationRecord
   enum search_engine: { google: 'google', bing: 'bing' }
   enum status: { in_progress: 'in_progress', completed: 'completed', failed: 'failed' }
 
+  belongs_to :user
```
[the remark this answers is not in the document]
hoangnguyen92dn: @Tuubz Since we want to add `current_user.search_results`, it requires to add `has_many`

app/controllers/api/v1/search_results_controller.rb
hoangnguyen92dn: @Tuubz Sounds good 👍 I will refactor `render_failed` to use `error_handler` instead

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -3,37 +3,46 @@
...
-      private
+      def index
+        search_result_list = current_user.search_results.order(:id)
+        pagy, paginated_results = paginated_resources_for(search_result_list)
 
-      def upload_form_params
-        params.permit(:search_engine, :csv_file)
+        search_result_serializer = SearchResultsSerializer.new(paginated_results, meta: meta_from_pagy(pagy))
+
+        render json: search_result_serializer, status: :ok
```
Tuubz, line 27:
```suggestion
        render_success(serializer: search_result_serializer, status: ok)
```

Tuubz (commented):
Rest LGTM 🔥

Tuubz (approved):

app/controllers/concerns/error_handler.rb
olivierobert: It is eithe rthe suffic `able` or the prefix `has`. See https://dev.to/mainstreet/how-to-name-rails-concerns-3m86

app/controllers/concerns/doorkeeper_authentication.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module DoorkeeperAuthentication
```
Tuubz, line 3: @hoangnguyen92dn  

Should this be renamed to `DoorkeeperAuthenticatable` or `HasDoorkeeperAuthentication`?  💭

hoangnguyen92dn: @Tuubz Following the article shared by Olivier, the `has` should be used in case we want to have a relation between 2 entity

> a has relation between the class where we will include the Concern to and a second entity

So I think we could name 2 concerns as below:

- error_handler -> error_handleable
- doorkeeper_authenticatication -> doorkeeper_authenticatable

💭

Tuubz:
@hoangnguyen92dn Conflicts 🙇

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -3,37 +3,46 @@
...
+      end
 
-        render json: { data: data, meta: { message: message } }, status: :created
+      def upload_form_params
+        params.permit(:search_engine, :csv_file)
       end
 
-      def render_failed(upload_form)
-        render json: { errors: upload_form.errors.full_messages }, status: :unprocessable_entity
+      def render_success(serializer:, status:)
```
hoangmirs, line 44: Do we need this method as it's just one line?
If we intend to use it in other controllers, should we move it into the ApplicationController instead?

app/controllers/concerns/pagination.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Pagination
+  extend ActiveSupport::Concern
+
```
hoangmirs, line 5: We can remove this line break as this is one group.

olivierobert (commented):
No major blockers but unsure about the way concerns are managed 💦

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -3,38 +3,43 @@
 module Api
   module V1
     class SearchResultsController < ApplicationController
-      before_action :doorkeeper_authorize!
+      include Paginable
+      include DoorkeeperAuthenticatable
```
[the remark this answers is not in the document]
hoangnguyen92dn: @olivierobert 🙈 Now I know why we call `concern` module in Ruby 🥲 just kidding 😅 

From my understanding after reading the article, the `has` should be used when concern has relation to other entities 💥 cc @Tuubz

olivierobert (approved):

hoangmirs (approved):

PR #48, opened 2023-07-11
[#10] [Backend] As a user, I can sign-in with e-mail and password

### LINKED ISSUE #10: [Backend] As a user, I can sign-in with e-mail and password
## Why

Allow a user to sign in to their account, once authenticated they can interact with the application.

## Acceptance Criteria

- Check if the user already exists, if not return an error.

- Create a new endpoint: `POST api/v1/sign_in`:

   - **Request**: E-mail, password, client ID & client secret

```
{
  "token_type": "bearer",
  "expires_in": 7200,
  "scope": "scopes"
}
```

## Design
N/A

## Resources
https://github.com/doorkeeper-gem/doorkeeper

Tuubz:
Close #10

## What happened 👀

- Added sign-in for email/password, can only be successful with prior sign-up of provider `email`
- Created `ValidClientConcern` so it can be reused across multiple controllers

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/b3d442bd-0f21-41b1-8d83-a9a1581cb0f2


app/controllers/api/v1/users/sessions_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+        before_action :ensure_valid_client
+
+        def create
+          user = User.find_by(email: params[:user][:email])
+          valid_provider = user&.provider == User.providers[:email]
+          valid_password = user&.valid_password?(params[:user][:password])
+
+          # TODO: Check if user has verified e-mail (#7)
+          if valid_provider && valid_password
+            render_success(user)
```
Tuubz, line 20: Note to self: Add `sign_in(user)`, otherwise `current_user` might return `nil` later on

hoangnguyen92dn:
@Tuubz I am curious if we could follow this [instruction from doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Resource-Owner-Password-Credentials-flow) to handle sign-in flow by username and password? 🤔 Then we don't need to handle stuff such as client_id validation and generating access_token.

On the Google sign-in flow, I had to generate an access token manually because we receive the callback via Omniauth to access the user profile for email to validate existing user, with `grant_type=password` flow, we could send request to `oauth/token` for sign-in 😄 Similar on [survey-web-api](https://documenter.getpostman.com/view/11835486/UVsPNPfU#deb6f64d-60a4-434b-9c83-215ad8a07568)

Tuubz:
> @Tuubz I am curious if we could follow this [instruction from doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Resource-Owner-Password-Credentials-flow) to handle sign-in flow by username and password? 🤔 Then we don't need to handle stuff such as client_id validation and generating access_token.
> 
> On the Google sign-in flow, I had to generate an access token manually because we receive the callback via Omniauth to access the user profile for email to validate existing user, with `grant_type=password` flow, we could send request to `oauth/token` for sign-in 😄 Similar on [survey-web-api](https://documenter.getpostman.com/view/11835486/UVsPNPfU#deb6f64d-60a4-434b-9c83-215ad8a07568)

Good find! I'll try to re-factor it 👍

Tuubz:
> @Tuubz I am curious if we could follow this [instruction from doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Resource-Owner-Password-Credentials-flow) to handle sign-in flow by username and password? 🤔 Then we don't need to handle stuff such as client_id validation and generating access_token.
> 
> On the Google sign-in flow, I had to generate an access token manually because we receive the callback via Omniauth to access the user profile for email to validate existing user, with `grant_type=password` flow, we could send request to `oauth/token` for sign-in 😄 Similar on [survey-web-api](https://documenter.getpostman.com/view/11835486/UVsPNPfU#deb6f64d-60a4-434b-9c83-215ad8a07568)

It's a good idea, but I can't make up my mind which approach is preferable in our case 💭 

- **Current approach**: `sessions_controller` & `omniauth_callbacks_controller` are aligned (easier to understand)
- **New approach**: Less code is required 

Perhaps there are some other advantages I'm missing out on... Thoughts? 🙏 

cc @olivierobert @hoangmirs

app/controllers/api/v1/users/sessions_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+module Api
+  module V1
+    module Users
+      class SessionsController < Devise::SessionsController
+        include ValidClientConcern
+
+        respond_to :json
+
+        def create
+          user = User.find_for_database_authentication(email: user_params[:email])
```
olivierobert, line 12: WHy using `find_for_database_authentication` instead of `find_by`?

app/controllers/api/v1/users/sessions_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+          valid_password = user&.valid_password?(user_params[:password])
+
+          return render_error([I18n.t('devise.failure.invalid', authentication_keys: :email)]) unless valid_password
```
olivierobert, line 15: This code is trying to be too smart IMO 😅 Prefer simpler:

```
user = User.find_by(email: user_params[:email])

if user&.valid_password?(user_params[:password])
  #...
else
  #...
end
```

Do not over-abuse `unless` 🧑‍🏫

app/controllers/concerns/valid_client_concern.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ValidClientConcern
```
olivierobert, line 3: Why create another concern? This could be placed in the existing doorkeeper-related concern.

Concerns should not be over-used.

Tuubz: I'll refactor this once after merging #13 🙏

hoangnguyen92dn:
@Tuubz Since we are building API application so session_controller would be confused IMO 😄 Just want to clarify if we could handle the feature already supported by Doorkeeper.

olivierobert (approved):

olivierobert:
@hoangnguyen92dn Even though technically REST API are stateless, so not holding sessions data, technically speaking the generated token create a temporary access. So no concerns in still naming it sessions_controller as it is consistent with most web apps. cc @Tuubz

hoangnguyen92dn (approved):

hoangmirs (approved):

Tuubz:
@hoangmirs @hoangnguyen92dn Could you please recheck my updates? 🙇 

- I refactored `ValidClientConcern` to be part of `HasDoorkeeperAuthentication`
- I also refactored the error handling to make use of `ErrorHandleable`

spec/requests/sessions_controller_spec.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Sessions', type: :request do
+  describe 'POST #create' do
+    context 'given a valid e-mail' do
+      it 'returns a successful response with JSON' do
+        application = Fabricate(:application)
+        allow(Doorkeeper::Application).to receive(:first).and_return(application)
```
hoangnguyen92dn, line 10: @Tuubz Now I see this stubbing is redundant 😄 We only need to fabricate the Application and that's all. Would you mind to update on other controller_spec 😅

hoangmirs: Yes, you just need to create an `application` as in your code, it will only use the first one 😄 
```suggestion
       Fabricate(:application)
       # or
       _application = Fabricate(:application)
```

Tuubz: @hoangmirs @hoangnguyen92dn Updated 🚀

hoangnguyen92dn (approved):
Go go 🚀

hoangmirs (approved):

PR #51, opened 2023-07-12
[#20] [Backend] As a user, I can filter which search results contain a specific URL

### LINKED ISSUE #20: [Backend] As a user, I can filter which search results contain a specific URL
## Why

A user can filter how many times a specific URL shows up in stored search results.

## Acceptance Criteria

- Use the existing endpoint to filter: `GET api/v1/search_results?filter[url_equals]={url}`

    -  **Request**: `filter[url_equals]={url}`

## Design
N/A

## Resources
N/A

toby-thanathip:
Close #20

## What happened 👀

Filter which search results have any specific URLs (`adwords_top_urls` & `non_adwords_urls`)

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/7bf95ad3-0035-4be3-b1cd-58ed2bb16ce6

### DIFF
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -22,7 +22,12 @@ module Api
       end
 
       def index
-        search_result_list = current_user.search_results.order(:id)
+        search_result_list = SearchResultsQuery.new(
+          current_user.search_results, {
+            url_equals: params.dig(:filter, :url_equals)
+          }
+        ).call
+
         pagy, paginated_results = paginated_resources_for(search_result_list)
 
         search_result_serializer = SearchResultsSerializer.new(paginated_results, meta: meta_from_pagy(pagy))
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -24,7 +24,7 @@ module Api
       def index
         search_result_list = SearchResultsQuery.new(
           current_user.search_results, {
-            url_equals: params.dig(:filter, :url_equals)
+            url_equals: filter_params[:url_equals]
           }
         ).call
 
@@ -47,6 +47,10 @@ module Api
 
       def upload_form_params
         params.permit(:search_engine, :csv_file)
+      end
+
+      def filter_params
+        params.fetch(:filter, {}).permit(:url_equals)
       end
     end
   end
--- a/app/queries/search_results_query.rb
+++ b/app/queries/search_results_query.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-CONDITION_URL_EQUALS = ':url = ANY (adwords_top_urls) OR :url = ANY (non_adwords_urls)'
-
 class SearchResultsQuery
   attr_reader :search_results, :filters, :sorts
+
+  CONDITION_URL_EQUALS = ':url = ANY (adwords_top_urls) OR :url = ANY (non_adwords_urls)'
 
   def initialize(search_results = nil, filters = {})
     @search_results = search_results || SearchResult.all
--- a/app/queries/search_results_query.rb
+++ b/app/queries/search_results_query.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-CONDITION_URL_EQUALS = '? = ANY (adwords_top_urls) OR ? = ANY (non_adwords_urls)'
+CONDITION_URL_EQUALS = ':url = ANY (adwords_top_urls) OR :url = ANY (non_adwords_urls)'
 
 class SearchResultsQuery
   attr_reader :search_results, :filters, :sorts
@@ -11,10 +11,16 @@ class SearchResultsQuery
   end
 
   def call
-    url = filters[:url_equals]
-
-    @search_results = @search_results.where(CONDITION_URL_EQUALS, url, url) if url.present?
+    @search_results = filter_by_url(filters[:url_equals])
 
     @search_results
+  end
+
+  def filter_by_url(url)
+    if url.present?
+      @search_results.where(CONDITION_URL_EQUALS, url: url)
+    else
+      @search_results
+    end
   end
 end
--- a/app/queries/search_results_query.rb
+++ b/app/queries/search_results_query.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+CONDITION_URL_EQUALS = '? = ANY (adwords_top_urls) OR ? = ANY (non_adwords_urls)'
+
+class SearchResultsQuery
+  attr_reader :search_results, :filters, :sorts
+
+  def initialize(search_results = nil, filters = {})
+    @search_results = search_results || SearchResult.all
+    @filters = filters
+  end
+
+  def call
+    url = filters[:url_equals]
+
+    @search_results = @search_results.where(CONDITION_URL_EQUALS, url, url) if url.present?
+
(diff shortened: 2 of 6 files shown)


app/controllers/api/v1/search_results_controller.rb
```diff
@@ -44,6 +46,15 @@ def upload_form_params
       def render_success(serializer:, status:)
         render json: serializer, status: status
       end
+
+      def filter_url_equals(search_result_list)
+        url = params.dig(:filter, :url_equals)
```
olivierobert, line 51: It would be more secure to use strong params:

```ruby
def filter_params
   params.requires(filter).permit(:url)
end 
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -44,6 +46,15 @@ def upload_form_params
...
+        if url
+          search_result_list.where('? = ANY (adwords_top_urls) OR ? = ANY (non_adwords_urls)', url, url)
+        else
+          search_result_list
```
olivierobert, line 55: This code is not elegant 🙈 Since there will be more filters, it would be best to create a query object: https://github.com/nimblehq/travelbook-web/blob/development/app/queries/travelogues_query.rb

app/controllers/api/v1/search_results_controller.rb
Tuubz: Hmmm, that would be nice, but the `filter` parameter isn't always required.

For example, we still want to be able to call `/api/v1/search_results` 

Having that said, can we keep it as it is? 🙏

olivierobert:
@Tuubz let's wait for Huey's PR to be merged to make more progress. It is not reviewable in its current state.

app/controllers/api/v1/search_results_controller.rb
hoangmirs: @Tuubz Could you try something like:
```ruby
def filter_params
  params.fetch(:filter, {}).permit(:url)
end 
```

app/queries/search_results_query.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+CONDITION_URL_EQUALS = '? = ANY (adwords_top_urls) OR ? = ANY (non_adwords_urls)'
```
hoangmirs: We can use the [`Placeholder Conditions`](https://guides.rubyonrails.org/active_record_querying.html#placeholder-conditions) to make it a bit clearer

```suggestion
CONDITION_URL_EQUALS = ':url = ANY (adwords_top_urls) OR :url = ANY (non_adwords_urls)'
```

And you can provide the name explicitly:

```ruby
@search_results.where(CONDITION_URL_EQUALS, url: url)
```

spec/queries/search_results_query_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SearchResultsQuery, type: :model do
+  describe '#call' do
+    context 'when given NO filters' do
```
hoangmirs: Should we use one of them `when`/`given`? We usually use `given:
```suggestion
    context 'given NO filters' do
```

app/queries/search_results_query.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(search_results = nil, filters = {})
+    @search_results = search_results || SearchResult.all
+    @filters = filters
+  end
+
+  def call
+    url = filters[:url_equals]
+
+    @search_results = @search_results.where(CONDITION_URL_EQUALS, url, url) if url.present?
```
olivierobert: It might be best to move the  logic to a private method so that additional queries can be added easily:

```suggestion
    @search_results = @search_results.filter_by_url(url: filters[:url_equals])

**Review by olivierobert — COMMENTED** [review]
Just a minor suggestion to make the code more future proof and also redable

File: app/queries/search_results_query.rb [hunk 6]
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+CONDITION_URL_EQUALS = ':url = ANY (adwords_top_urls) OR :url = ANY (non_adwords_urls)'
```
hoangnguyen92dn: This condition should be inside the class? 🤔 Reference on [Compass](https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#convention-31)

spec/queries/search_results_query_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SearchResultsQuery, type: :model do
```
hoangnguyen92dn, line 5: Not sure if the type is `model`? 🤔

hoangnguyen92dn:
Please be careful when rebasing this PR, since I changed the serializer name on #18 🙏

spec/queries/search_results_query_spec.rb
toby-thanathip: This should be correct: https://github.com/nimblehq/travelbook-web/blob/development/spec/queries/travelogues_query_spec.rb 💪

olivierobert (approved):

spec/requests/search_results_controller_spec.rb
```diff
@@ -79,6 +79,28 @@
       end
     end
 
+    context 'given a valid params with url_equals filter' do
+      it 'returns a successful response with JSON' do
+        user = Fabricate(:user)
+        Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['wwww.nimblehq.co'])
+        Fabricate(:search_result, user_id: user.id, adwords_top_urls: ['wwww.nimblehq.co'])
+        Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['wwww.nimblehq.co/compass'])
+        Fabricate(:search_result, user_id: user.id, adwords_top_urls: ['wwww.nimblehq.co/compass'])
```
olivierobert, line 92: After reviewing https://github.com/nimblehq/ic-rails-huey-toby/pull/53/files#diff-cbee06c3149f3c01bef664c25f0d4e6077787d3a0eff949d8f249d4dc39fc9cc, with hinsight, it becomes hard to read if we don't add variables names or some comment for this fabricated data. It would be best to do the following:

```ruby
# Expected matches
Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['www.nimblehq.co'])
Fabricate(:search_result, user_id: user.id, adwords_top_urls: ['www.nimblehq.co'])

# Non-matches
Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['www.nimblehq.co/compass'])
Fabricate(:search_result, user_id: user.id, adwords_top_urls: ['www.nimblehq.co/compass'])
```

hoangmirs (approved):

spec/queries/search_results_query_spec.rb
hoangnguyen92dn: @Tuubz Yeah... I am confused about this, seems we need the `:type` when having some config before each cases, on ewapayroll, they don't have type for query spec files 😄 

https://github.com/nimblehq/ewa-payroll-web/blob/049a5896bc4e172cfee496e8e61f58af4b1e4e38/spec/queries/holiday_query_spec.rb#L4

hoangnguyen92dn (approved):

PR #53, opened 2023-07-14
[#17] [Backend] As a user, I can filter which search results contain a word in Adwords URLs

### LINKED ISSUE #17: [Backend] As a user, I can filter which search results contain a word in Adwords URLs
## Why

A user can filter how many URLs contain a specific word in the Adwords URLs

## Acceptance Criteria

- Use the existing endpoint to filter: `GET api/v1/search_results?filter[adwords_url_contains]={word}`

    -  **Request**: `filter[adwords_url_contains]={word}`

## Design
N/A

## Resources
N/A

toby-thanathip:
Close #17

## What happened 👀

Add filter to return search results which have a word containing inside `adwords_top_urls`

## Insight 📝

The `LIKE` operator which is used to check if a string contains a word, can't be used on an array. 

As a solution, I convert the array of strings to a string first, before using the `LIKE` operator
 
**Note:** It's crucial to surround the word with `%`  inside the query
 
## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/18277915/f77d24eb-e28d-4bd6-b139-e49b58e7e7c2

### DIFF
--- a/spec/queries/search_results_query_spec.rb
+++ b/spec/queries/search_results_query_spec.rb
@@ -9,10 +9,10 @@ RSpec.describe SearchResultsQuery, type: :model do
         user = Fabricate(:user)
 
         # Expected matches
-        Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['wwww.nimblehq.co'])
-        Fabricate(:search_result, user_id: user.id, adwords_top_urls: ['wwww.nimblehq.co'])
-        Fabricate(:search_result, user_id: user.id, non_adwords_urls: ['wwww.nimblehq.co/compass'])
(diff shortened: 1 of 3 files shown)


spec/queries/search_results_query_spec.rb
```diff
@@ -7,10 +7,10 @@
     context 'given NO filters' do
       it 'returns all search results' do
         user = Fabricate(:user)
```
olivierobert, line 9: Revisit this setup based on https://github.com/nimblehq/ic-rails-huey-toby/pull/51/files#r1266122433

olivierobert (approved):

hoangmirs (approved):

hoangnguyen92dn (approved):

PR #57, opened 2023-07-19
[#21][Backend] As a user, I can filter which search results contain a specific amount of characters

### LINKED ISSUE #18: [Backend] As a user, I can see the details of a search result

### LINKED ISSUE #21: [Backend] As a user, I can filter which search results contain a specific amount of characters
## Why

A user can filter which search results have URLs in stored search results with 2 or more `/` or 1 or more `>`.

## Acceptance Criteria

- Use the existing endpoint to filter: `GET api/v1/search_results?filter[url_contains_at_least_two]={character}`

    -  **Request**: `filter[url_contains_at_least_two]={character}`
    

- Use the existing endpoint to filter: `GET api/v1/search_results?filter[url_contains_at_least_one]={character}`

    -  **Request**: `filter[url_contains_at_least_one]={character}`
    
## Design
N/A

## Resources
N/A

hoangnguyen92dn:
- Close #21 

## What happened 👀

- Use the existing endpoint to filter: ~~`GET api/v1/search_results?filter[url_contains_at_least_two]={character}`~~ `GET api/v1/search_results?filter[url_contains]={character}&filter[match_count]={number}`

    -  **Request**: ~~`filter[url_contains_at_least_two]={character}`~~ `filter[url_contains]={character}&filter[match_count]={number}`

## Insight 📝

It's so tricky to look for `search_result` records which have URL containing a specific character by using the query. The alternative approach is filtering the urls match condition and then querying by the ids.

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-huey-toby/assets/6950766/2874dee2-d165-460e-9616-7606884a875f

### DIFF
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -46,8 +46,8 @@ module Api
           current_user.search_results, {
             url_equals: filter_params[:url_equals],
             adwords_url_contains: filter_params[:adwords_url_contains],
-            url_contains_at_least_one: filter_params[:url_contains_at_least_one],
-            url_contains_at_least_two: filter_params[:url_contains_at_least_two]
+            url_contains: filter_params[:url_contains],
+            match_count: filter_params[:match_count]
           }
         ).call
       end
@@ -65,8 +65,8 @@ module Api
       end
 
       def filter_params
-        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains, :url_contains_at_least_one,
-                                         :url_contains_at_least_two)
+        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains, :url_contains,
+                                         :match_count)
       end
 
       def upload_meta(search_result_list:)
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -45,7 +45,9 @@ module Api
         SearchResultsQuery.new(
           current_user.search_results, {
             url_equals: filter_params[:url_equals],
-            adwords_url_contains: filter_params[:adwords_url_contains]
+            adwords_url_contains: filter_params[:adwords_url_contains],
+            url_contains_at_least_one: filter_params[:url_contains_at_least_one],
+            url_contains_at_least_two: filter_params[:url_contains_at_least_two]
           }
         ).call
       end
@@ -63,7 +65,8 @@ module Api
       end
 
       def filter_params
-        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains)
+        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains, :url_contains_at_least_one,
+                                         :url_contains_at_least_two)
       end
 
       def upload_meta(search_result_list:)
--- a/app/controllers/api/v1/search_results_controller.rb
+++ b/app/controllers/api/v1/search_results_controller.rb
@@ -65,8 +65,9 @@ module Api
       end
 
       def filter_params
-        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains, :url_contains,
-                                         :match_count)
+        params
+          .fetch(:filter, {})
+          .permit(:url_equals, :adwords_url_contains, :url_contains, :match_count)
       end
 
       def upload_meta(search_result_list:)
--- a/app/queries/search_results_query.rb
+++ b/app/queries/search_results_query.rb
@@ -7,20 +7,22 @@ class SearchResultsQuery
   CONDITION_ADWORDS_URL_CONTAINS = "array_to_string(adwords_top_urls, ', ') LIKE :word"
 
   def initialize(search_results = nil, filters = {})
-    @search_results = search_results || SearchResult.all
+    @search_results = search_results || SearchResult.order(:id).all
     @filters = filters
   end
 
   def call
     @search_results = filter_by_url(filters[:url_equals])
     @search_results = filter_adwords_url_contains(filters[:adwords_url_contains])
+    @search_results = filter_urls_contains_at_least(filters[:url_contains_at_least_one], 1)
+    @search_results = filter_urls_contains_at_least(filters[:url_contains_at_least_two], 2)
 
     @search_results
   end
 
   def filter_by_url(url)
     if url.present?
-      @search_results.where(CONDITION_URL_EQUALS, url: url)
+      @search_results.where(CONDITION_URL_EQUALS, url: url).order(:id)
     else
       @search_results
     end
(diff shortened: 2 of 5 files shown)


app/queries/search_results_query.rb
```diff
@@ -7,30 +7,51 @@ class SearchResultsQuery
   CONDITION_ADWORDS_URL_CONTAINS = "array_to_string(adwords_top_urls, ', ') LIKE :word"
 
   def initialize(search_results = nil, filters = {})
-    @search_results = search_results || SearchResult.all
+    @search_results = search_results || SearchResult.order(:id).all
```
toby-thanathip: If I'm not mistaken, by default it's already ordered on `:id`

Did I miss something? 🙏

app/queries/search_results_query.rb
```diff
@@ -7,30 +7,51 @@ class SearchResultsQuery
...
+      @search_results.where(id: search_result_ids).order(:id)
+    else
+      @search_results
+    end
+  end
+
+  private
+
+  def at_least_character_count?(urls, character, count)
+    urls.any? { |url| url.scan(character).count >= count }
```
toby-thanathip: The test still run successfully with this update 🚀 

```suggestion
    urls.any? { |url| url.count(character) >= count }
```

app/queries/search_results_query.rb
```diff
@@ -7,30 +7,51 @@ class SearchResultsQuery
...
     else
       @search_results
     end
   end
+
+  def filter_urls_contains_at_least(character, count = 1)
+    if character.present?
+      search_result_ids = @search_results.select do |search_result|
+        at_least_character_count?(search_result.adwords_top_urls, character, count) ||
+          at_least_character_count?(search_result.non_adwords_urls, character, count)
```
toby-thanathip: Optional: This is easier to read IMO, but I'm fine if you keep the current one too 🙏 

```suggestion
        adwords_top_urls_contains = urls.any? { |url| url.count(character) >= count }
        non_adwords_urls_contains = urls.any? { |url| url.count(character) >= count }
        adwords_top_urls_contains || non_adwords_urls_contains
```

toby-thanathip (commented):
🔥

app/queries/search_results_query.rb
hoangnguyen92dn: @Tuubz I noticed the list does not sorted by id that's why I added the `order`

image

app/queries/search_results_query.rb
hoangnguyen92dn: @Tuubz Can you please check this case on your side? 🤔 When using `count` instead of `scan`, it returns all the records.

image

image

toby-thanathip: Hmmmm, interesting...

Seems like they function differently when _more than one character_ is given 💭 

Screen Shot 2023-07-20 at 09 17 40

Cool, let's go with your approach 🚀

app/queries/search_results_query.rb
toby-thanathip: I see, let's go with your approach 👍  

However, would it be better to call `.order(:id)` only once on line `20`?

app/queries/search_results_query.rb
toby-thanathip: I believe this is required to prevent it from throwing an exception if `urls` is `nil`

```suggestion
    urls&.any? { |url| url.scan(character).count >= count }
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -45,7 +45,9 @@ def create_search_results_query
         SearchResultsQuery.new(
           current_user.search_results, {
             url_equals: filter_params[:url_equals],
-            adwords_url_contains: filter_params[:adwords_url_contains]
+            adwords_url_contains: filter_params[:adwords_url_contains],
+            url_contains_at_least_one: filter_params[:url_contains_at_least_one],
+            url_contains_at_least_two: filter_params[:url_contains_at_least_two]
```
olivierobert: These filters are not extensible 🙈 How about breaking them into two instead:

```
url_contains: filter_params[:url_contains],
match_count: filter_params[:match_count],
```

So we could search for 1, 2, ♾️ number of occurrences 💡

app/queries/search_results_query.rb
hoangnguyen92dn: @Tuubz I'd prefer to keep the current one 😄 So we would split to logic to filter URLs with matched character count.

toby-thanathip (approved):
Shall we also update the PR description? 🚀

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -63,7 +65,8 @@ def upload_form_params
       end
 
       def filter_params
-        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains)
+        params.fetch(:filter, {}).permit(:url_equals, :adwords_url_contains, :url_contains,
```
olivierobert: Thsi alignment is weird 🙈 Either all attributes should be on one one line or split onto its own line:

```ruby
params
   .fetch(:filter, {})
   .permit(:url_equals, :adwords_url_contains, :url_contains, :match_count)
```

Or:

```ruby
params .fetch(:filter, {}).permit(:url_equals, 
                                  :adwords_url_contains, 
                                  :url_contains, 
                                  :match_count)
```

app/queries/search_results_query.rb
```diff
@@ -33,4 +34,23 @@ def filter_adwords_url_contains(word)
       @search_results
     end
   end
+
+  def filter_urls_contains_at_least(character, match_count = 1)
```
olivierobert: For consistency, the parameter could be named `word` 💡 It is okay to have your own take of the requirements. You need to build the feature in a way that makes sense.

app/queries/search_results_query.rb
```diff
@@ -33,4 +34,23 @@ def filter_adwords_url_contains(word)
...
+    if character.present?
+      search_result_ids = @search_results.select do |search_result|
+        at_least_character_count?(search_result.adwords_top_urls, character, match_count) ||
+          at_least_character_count?(search_result.non_adwords_urls, character, match_count)
```
olivierobert: Both arrays would be joined so that the `OR` can be removed:

```ruby
search_result.adwords_top_urls + search_result.non_adwords_urls
```

spec/queries/search_results_query_spec.rb
```diff
@@ -61,5 +61,65 @@
         expect(search_results_query.search_results.count).to eq 1
       end
     end
+
+    context 'given filter_urls_contains filter' do
+      context 'given filter_match_count at least one filter' do
```
olivierobert: These tests do not make sense syntactically (missing verb) and logically (why searching for 1 and 2 occurrences since match_count can be dynamic 🙃 ). Since match_count can be omitted, the following tests would make more sense:

```
context 'given filter_urls_contains filter' do
   context 'given match_count filter is nil' do
      it 'returns only the search results which contain specific  word at least once' do
        #...
     end
   end

   context 'given match_count filter is NOT nil' do
      it 'returns only the search results which contain a specific word at least for the define occurrence match count'  do
        #...
     end
   end

   it 'returns empty when the urls do NOT match a specific word' do
     #...
  end
end
```

olivierobert (changes requested):

spec/queries/search_results_query_spec.rb
```diff
@@ -61,5 +61,65 @@
...
+          search_results_query = described_class.new(nil, { url_contains: '/' })
+
+          search_results_query.call
+
+          expect(search_results_query.search_results.count).to eq 2
+        end
+      end
+
+      context 'given match_count filter is NOT nil' do
+        it 'returns only the search results which contain a specific word at least for the define occurrence match count' do
```
hoangmirs:
```suggestion
        it 'returns only the search results which contain a specific word at least for the defined occurrence match count' do
```

hoangmirs (approved):

olivierobert (approved):
Just one minor dypo

hoangnguyen92dn:
@toby-thanathip Go go 🚀

